Issue #15, opened 2026-07-21
ironfix-store: get_range panics via public API (BTreeMap inverted range, end+1 overflow) and fabricates MsgType::Heartbeat; engine never uses MessageStore at all


joaquinbejar:
## Overview

MemoryStore::get_range aborts the process on begin > end (std BTreeMap::range panics, reachable from a plain trait call) and overflows end+1 in the end=0 error branch; retrieved messages are rebuilt with a fabricated MsgType::Heartbeat and lose their sequence keys, making a correct resend impossible. Meanwhile the engine declares ironfix-store as a dependency but never touches it — every ResendRequest is answered with a blanket GapFill, so missed application messages are permanently unrecoverable, and the crate docs advertise a FileStore that does not exist.

**Priority: P1** — from the full-workspace audit of 2026-07-21 (multi-agent adversarially-verified review, cross-checked against `doc/code_audit_2026-07-21.md`; findings marked `[source: codex-audit]` originate there).

## Context

Crates / areas touched: `ironfix-store`, `ironfix-engine`

### 1. get_range panics inside BTreeMap::range when begin > end (finding 123)
`ironfix-store/src/memory.rs:97` (94-101)

`let end = if end == 0 { u64::MAX } else { end }; let result: Vec<OwnedMessage> = messages.range(begin..=end)...` — no begin<=end validation before `.range()`, and std BTreeMap::range is documented to panic when start > end. get_range(5, 3) aborts the process under panic=abort from a plain trait-method call. The `begin <= end` guard at line 101 protects only the error-construction branch, after the panic would already have fired. (Corrects the earlier claim that begin>end returns silent Ok.) [source: codex-audit]

**Fix:** before `.range()`, return a typed StoreError (or Ok(vec![])) when begin > end after normalization; tests for get_range(5,3).

### 2. `end + 1` overflows when end=0 ('infinity') finds no messages; begin > end semantics undefined (finding 68)
`ironfix-store/src/memory.rs:103`

end=0 maps to u64::MAX; the empty-result error path computes `range: begin..end + 1` — overflow panic in debug, wrapped nonsense range in release — reachable via the documented API (get_range(5,0) on an empty store). No test covers end=0 or RangeNotAvailable.

**Fix:** inclusive (begin,end) in the error payload or saturating_add; typed error for begin>end; tests for all three paths.

### 3. get_range fabricates MsgType::Heartbeat and drops seq association (finding 69)
`ironfix-store/src/memory.rs:98`

[...]

Issue #10, opened 2026-07-21
ironfix-session: heartbeat logic self-destructs on HeartBtInt=0 (remotely triggerable), ignores live traffic after a TestRequest, and hard-codes the grace period


joaquinbejar:
## Overview

HeartbeatManager with a zero interval floods heartbeats, fires TestRequest after 1s, and is instantly timed out — and the engine adopts the counterparty's Logon 108 verbatim with no bounds, so a peer can trigger the storm (108=0) or disable dead-peer detection (108=u64::MAX). is_timed_out ignores all traffic received after a TestRequest, so demonstrably live sessions get torn down. The disconnect-decision branches have no unit tests and the 1-second grace is hard-coded.

**Priority: P1** — from the full-workspace audit of 2026-07-21 (multi-agent adversarially-verified review, cross-checked against `doc/code_audit_2026-07-21.md`; findings marked `[source: codex-audit]` originate there).

## Context

Crates / areas touched: `ironfix-session`, `ironfix-engine`

### 1. HeartBtInt=0 unguarded end to end — heartbeat storm then instant timeout, remotely reachable (finding 49)
`ironfix-session/src/heartbeat.rs:79`, `ironfix-engine/src/initiator.rs:287-295`

With interval==ZERO, `self.last_sent.elapsed() >= self.interval` is always true (flood every tick, burning seqnums), should_send_test_request fires after 1s, and is_timed_out() is instantly true — the session self-destructs. Remotely reachable: initiator.rs:287-295 installs the Logon ack's 108 verbatim (`*heartbeat = HeartbeatManager::new(Duration::from_secs(secs))`) with no lower/upper bound, so 108=0 triggers the storm+disconnect and 108=u64::MAX disables dead-peer detection entirely. Untrusted input driving timer arithmetic with no ceiling.

**Fix:** model HeartBtInt=0 explicitly (all three predicates false, per FIX semantics) and clamp/reject out-of-range counterparty 108 (e.g. 1..=3600 or tolerance vs requested) in the engine; unit tests for zero interval.

### 2. is_timed_out ignores all traffic received after a TestRequest (finding 50)
`ironfix-session/src/heartbeat.rs:102`

[...]

PR #25, opened 2026-07-21
fix(engine): make inbound session handling conform to the FIX spec

### LINKED ISSUE #9: ironfix-engine: inbound protocol conformance — GapFill conflated with Reset (real gaps silently masked), ResendRequest reply over-advances peers, missing OrigSendingTime, FIXT 5.0 wire-nonconformant, no CompID validation, wrapping seqnum arithmetic
## Overview

The Initiator's inbound session handling deviates from the FIX spec in ways that lose messages and desynchronize live sessions: SequenceReset-GapFill is applied without validating its own MsgSeqNum (skipping potentially ExecutionReports forever), ResendRequest replies ignore EndSeqNo and default a missing BeginSeqNo to 1, the only resent frame omits required OrigSendingTime(122), FIX 5.0 sessions send the wrong BeginString, inbound 141=Y and CompIDs are never checked, and every sequence increment uses the silently-wrapping variants CLAUDE.md marks red. None of these recovery paths has an end-to-end test.

**Priority: P0** — from the full-workspace audit of 2026-07-21 (multi-agent adversarially-verified review, cross-checked against `doc/code_audit_2026-07-21.md`; findings marked `[source: codex-audit]` originate there).

## Context

Crates / areas touched: `ironfix-engine`, `ironfix-session`

### 1. Inbound SequenceReset conflates GapFill and Reset modes — real gaps silently masked (finding 66)
`ironfix-engine/src/initiator.rs:608`

Per FIX only Reset mode ignores MsgSeqNum; a GapFill (123=Y) participates in normal sequencing. The handler runs before validate_incoming — comment: 'SequenceReset (including GapFill) jumps the target expectation regardless of its own MsgSeqNum.' A GapFill with 34=10 while expecting 5 and 36=12 jumps the target to 12: messages 5..9 (potentially ExecutionReports) are skipped and never resent. A NewSeqNo below expected is only warned about instead of Rejected (reason 5), and the from_admin verdict is discarded (`let _ =`), unlike every other admin type.

**Fix:** split on tag 123: for 123=Y validate 34 first (Gap → ResendRequest without applying 36; TooLow+43=Y → drop; Ok → apply); Reset mode alone bypasses 34; Reject (reason 5) on decreasing NewSeqNo; honor the from_admin verdict.

### 2. ResendRequest reply: missing BeginSeqNo defaults to 1, EndSeqNo(16) ignored (finding 77)
`ironfix-engine/src/initiator.rs:692`

`let begin_seq = raw.get_field_str(7)...unwrap_or(1);` (forbidden silent default) then `sequence_reset_gap_fill(begin_seq, next_sender_seq)` — tag 16 is never read. A bounded request (7=5,16=10) while we've sent to 100 gets a GapFill jumping the peer's expectation to 100, skipping everything past EndSeqNo and desynchronizing the session. begin_seq >= next_sender_seq produces a nonsense GapFill whose 36 is below its own 34.

**Fix:** Reject (reason 1, RefTagID=7) on absent/invalid 7 and reason 5 on out-of-range; bound NewSeqNo to EndSeqNo+1 for bounded requests; reactor tests for malformed/out-of-range ResendRequests.

### 3. GapFill sent with PossDupFlag=Y but without OrigSendingTime(122) (finding 76)
`ironfix-engine/src/wire.rs:151`

FIX requires 122 on any message with 43=Y and doc/fix_operations.md:152 mandates it for resends; the only resent frame the engine ever emits omits it — grep: no 122 anywhere in ironfix-engine/src. QuickFIX-based and strict venue acceptors reject with 'required tag missing' exactly on the recovery path.

**Fix:** stamp 122 whenever poss_dup is true (current time per the spec's data-unavailable clause absent a store); extend test_gap_fill_frame to assert 122.

### 4. 5.0/FIXT sessions wire-nonconformant: FIX.5.0* in tag 8, no 1137/1128 (finding 80)
`ironfix-engine/src/wire.rs:35`

static_begin_string interns "FIX.5.0"/"FIX.5.0SP1"/"FIX.5.0SP2" verbatim into tag 8, but every 5.0 variant MUST send BeginString=FIXT.1.1 with the variant in ApplVerID — venues reject 8=FIX.5.0 outright. Logon never emits DefaultApplVerID(1137), application_message never stamps ApplVerID(1128) (both required per doc/fix_operations.md:675-678); grep for 1137/1128 in engine src returns nothing. The dictionary already knows the right answer (Version::begin_string).

**Fix:** route the configured version through ironfix_dictionary Version semantics (tag 8 = transport BeginString), add 1137 to Logon and 1128 to app frames for FIXT sessions — or remove the 5.0 arms and the version-matrix rows until supported.

### 5. Logon ack ResetSeqNumFlag(141)=Y not honored inbound (finding 81)
`ironfix-engine/src/initiator.rs:299`

The handshake validates the ack's 34 but never inspects 141; a peer-driven reset (141=Y, 34=1) against a continuity-seeded local target > 1 yields TooLow and aborts the handshake, though FIX requires both sides to reset. Only outbound reset_on_logon is implemented.

**Fix:** check the ack for 141=Y before validating 34; reset both counters to 1 and clear pending resend.

### 6. Inbound CompID (49/56) and SendingTime (52) never validated — reject reasons 9/10 unimplemented (finding 71)
`ironfix-engine/src/initiator.rs:585`

on_frame dispatches purely on MsgType + 34; 49/56/52 are never read. A cross-wired connection is silently accepted, advancing sequence state against the wrong counterparty and delivering another session's execution reports. doc/fix_operations.md enumerates reasons 9 and 10.

**Fix:** compare 49/56 (and 50/57 when configured) post-decode; Reject reason 9 + Logout on mismatch; optional SendingTime window (reason 10); stub-acceptor tests.

### 7. Engine uses silently-wrapping fetch_add seqnum methods at every call site (finding 51)
`ironfix-engine/src/initiator.rs:224`, `ironfix-session/src/sequence.rs:107`

`SeqNum::new(self.next_sender_seq.fetch_add(1, Ordering::SeqCst))` — doc admits 'wraps silently on u64 overflow' — used at ~16 engine sites (initiator.rs 224, 275, 301, 319, 499, 514, 559, 567, 635, 644, 652, 671, 684, 730, 745) while try_allocate_sender_seq/try_increment_target_seq and SequenceExhausted exist unused. Counters are seedable arbitrarily via with_initial/set_* (e.g. from a corrupted store). CLAUDE.md marks wrapping seqnum arithmetic red.

**Fix:** switch the engine to the try_ variants, surfacing SequenceExhausted as session teardown; #[deprecated] (or privatize) the wrapping methods, removing at the next major.

### 8. No end-to-end test covers sequence recovery or teardown (finding 73)
`ironfix-engine/tests/initiator_tests.rs:470`

Only gap DETECTION is tested. Untested reachable branches: inbound 35=4 (GapFill/Reset application, backward guard), inbound 35=2 → GapFill reply fields (where the missing-122 and ignored-EndSeqNo bugs live), PossDup duplicate drop, seq-too-low without 43=Y → Logout+close, gapped Logon ack → pre-reactor ResendRequest, 108 renegotiation, logout-ack timeout, drop-all-handles auto-logout, from_admin rejection → session Reject, undecodable/34-less frames, reset_on_logout/disconnect, with_initial_sequences seeding. grep: no venue_msg("4"), no venue_msg("2"), no (43,"Y") fixture.

**Fix:** extend the stub-acceptor/RecordingApp harness with one test per branch listed.

## Tasks

- [ ] Inbound SequenceReset conflates GapFill and Reset modes — real gaps silently masked (finding 66)
- [ ] ResendRequest reply: missing BeginSeqNo defaults to 1, EndSeqNo(16) ignored (finding 77)
- [ ] GapFill sent with PossDupFlag=Y but without OrigSendingTime(122) (finding 76)
- [ ] 5.0/FIXT sessions wire-nonconformant: FIX.5.0* in tag 8, no 1137/1128 (finding 80)
- [ ] Logon ack ResetSeqNumFlag(141)=Y not honored inbound (finding 81)
- [ ] Inbound CompID (49/56) and SendingTime (52) never validated — reject reasons 9/10 unimplemented (finding 71)
- [ ] Engine uses silently-wrapping fetch_add seqnum methods at every call site (finding 51)
- [ ] No end-to-end test covers sequence recovery or teardown (finding 73)

## Acceptance criteria

- [ ] Inbound GapFill (123=Y) with a gapped MsgSeqNum triggers ResendRequest without applying NewSeqNo; TooLow+43=Y is dropped; only Reset mode (no 123 or 123=N per spec reading) bypasses 34 validation — each covered by a stub-acceptor test
- [ ] A decreasing NewSeqNo is answered with session Reject reason 5; the from_admin verdict for SequenceReset is honored like other admin types
- [ ] ResendRequest with missing/invalid tag 7 is Rejected (reason 1, RefTagID=7); a bounded request never produces a GapFill whose 36 exceeds EndSeqNo+1; out-of-range requests Reject with reason 5
- [ ] Every frame carrying 43=Y also carries 122; test_gap_fill_frame asserts it
- [ ] For Fix50/SP1/SP2 sessions tag 8 is FIXT.1.1, Logon carries 1137, and app messages carry 1128 — or the 5.0 arms and doc matrix rows are removed with an explicit typed error; either way covered by tests
- [ ] Logon ack with 141=Y resets both counters to 1 before 34 validation (test with continuity-seeded counters)
- [ ] Inbound CompID mismatch produces Reject reason 9 + Logout in a stub-acceptor test; SendingTime window behavior is decided and documented
- [ ] All engine sequence increments go through checked try_ variants; SequenceExhausted tears the session down with a typed error; wrapping variants are deprecated/private
- [ ] initiator_tests covers each recovery/teardown branch enumerated in finding 73

joaquinbejar:
## Summary

Makes the `Initiator`'s inbound session handling conform to the FIX specification. Eight defects from the 2026-07-21 audit, all reachable from a counterparty, several of which lose messages or desynchronise a live session.

Closes #9. Third and last of the P0 set; #23 and #24 are already merged, so this branches straight off `main` and the diff is only its own.

## What was wrong

**1. SequenceReset conflated GapFill with Reset.** Only Reset mode may ignore MsgSeqNum — a GapFill participates in normal sequencing. The old handler jumped the inbound expectation regardless, so a GapFill with `34=10` while expecting 5 skipped messages 5..9, potentially ExecutionReports, permanently. Now `123=Y` validates MsgSeqNum first: a gapped fill triggers a ResendRequest **without** applying NewSeqNo, a too-low duplicate with `43=Y` is dropped, and a decreasing NewSeqNo is Reject reason 5. The `from_admin` verdict is honoured instead of discarded.

**2. The ResendRequest reply ignored the request.** BeginSeqNo silently defaulted to 1 when absent, and EndSeqNo was never read — so a bounded request (`7=5, 16=10`) from a peer we had sent 100 messages to was answered by gap-filling to 100, jumping it past everything it asked to keep. Both tags are now validated (reason 1 with the right RefTagID), `16=0` means infinity, out-of-range is reason 5, and NewSeqNo is bounded by `min(EndSeqNo + 1, next_sender)`.

The reply also now carries **BeginSeqNo as its own MsgSeqNum** and allocates no sender number, because a SequenceReset-GapFill occupies the range it replaces. Verified against FIX 4.4 Vol 2 and QuickFIX/J's `generateSequenceReset`.

**3. PossDupFlag without OrigSendingTime.** The only resent frame the engine emits carried `43=Y` with no tag 122, which strict acceptors reject on exactly the recovery path. 122 is now stamped from the same flag, so it cannot be omitted. With no store there is no original sending time to recover, so the current SendingTime is used — which is what QuickFIX/J does for generated gap fills.

**4. FIX 5.0 sessions were wire-nonconformant.** `FIX.5.0*` went into tag 8 verbatim; every 5.0 variant must send `BeginString=FIXT.1.1` with the variant in ApplVerID. Now FIXT.1.1 with DefaultApplVerID(1137) on the Logon and ApplVerID(1128) on application messages only, in its correct standard-header position.

A version the engine cannot frame conformantly is now **refused before dialling** with `EngineError::UnsupportedVersion`: an unknown string (previously passed through onto the wire, which the coding rules forbid) or `FIXT.1.1` on its own, which names a transport version but no application version for the *required* 1137. Neither is defaulted to a guess.

**5. Inbound ResetSeqNumFlag was ignored.** A peer-driven reset (`141=Y, 34=1`) against continuity-seeded counters read as fatally too low and aborted the handshake. Now honoured before MsgSeqNum validation — and required to arrive under `MsgSeqNum = 1`, since a reset and the number carrying it have to describe the same stream.

**6. No inbound identity validation.** Tags 49/56 were never read, so a cross-wired connection established a session and advanced sequence state against the wrong counterparty — delivering another session's execution reports. Now compared (reversed) on both the Logon ack and every reactor frame, with 50/57 when configured; a mismatch is Reject reason 9 followed by a Logout.

**7. Heartbeat bookkeeping ran before validation**, so foreign or invalid traffic kept the connection alive. It now runs after identity validation.

**8. Every sequence increment wrapped silently.** ~15 call sites used `fetch_add`-based methods documented as wrapping at `u64::MAX`, which corrupts a live session. All now use the checked variants; exhaustion surfaces as a typed teardown, and the wrapping pair is `#[deprecated]`.

## A critical regression this PR's own first draft introduced

Worth stating plainly. The new `from_admin` rejection path returned **without consuming the rejected message's MsgSeqNum**, even when the GapFill was in sequence and therefore occupied that number. Adversarial review reproduced the consequence against a live `Initiator`:

```
venue: one in-sequence GapFill the application rejects, then well-behaved
       ExecutionReports at 3, 4, 5, 6

before the fix: reports delivered = []        next_target_seq = 2
                is_closed = false   is_timed_out = false   on_logout never fired
after the fix:  reports delivered = [E3, E4, E5, E6]   next_target_seq = 7
```

Four ExecutionReports from a well-behaved venue lost after a single admin rejection, with **no signal of any kind** to the application — the expectation parks on the unconsumed number, every later message looks gapped, gets deduplicated against the outstanding ResendRequest, and is dropped. All 33 tests were green while this was live.

Ironically the same function guards against exactly this hazard thirty lines below, with a comment naming it. The fix consumes the number whenever a rejected fill was in sequence, on both rejection paths. The regression test is verified to fail without the fix and pass with it.

## What this PR deliberately does **not** implement

- **Replay from a `MessageStore`** — that is #15. The gap-fill is now correct and bounded, but no application message is ever actually replayed, so mandate items 1, 2 and 4 of the spec's Resend Request section remain unmet. The Phase 1 checkbox stays unticked and says why.
- **SendingTime accuracy checking** (reject reason 10). A clock-skew tolerance is a policy decision needing its own typed knob and a defensible default; guessing one would be worse than declaring the gap.
- **A ceiling on inbound `NewSeqNo`.** A peer can jump the expectation to `u64::MAX`, after which the next message exhausts the counter and the session is torn down with a typed error. The spec offers no principled limit, and a clean teardown beats an invented one. (This is still an improvement: the old code wrapped silently to 0 and carried on, which was real corruption.)
- **Heartbeat conformance.** Adversarial review confirmed two live defects in `ironfix-session::HeartbeatManager` — `HeartBtInt = 0` makes the session flood the peer and time itself out, and a pending TestRequest is cleared only by a Heartbeat echoing tag 112, so a venue answering without it is disconnected while demonstrably alive. Both are issue #10, out of scope here — but "Heartbeat / Test Request" has been **unticked** from the Phase 1 list rather than left claiming to work.
- **Positional CompID enforcement.** Lookup is by tag over the whole message, so CompIDs appearing outside the standard header still satisfy the check. Documented on `PeerIdentity::validate`; fixing it needs positional information the decoder does not retain.

The workspace is already at 0.4.0. `cargo semver-checks --release-type minor` reports `enum_variant_added` on `EngineError`, i.e. major required — satisfied.

- `EngineError` gains `#[non_exhaustive]` (matching every sibling error enum after #23/#24) plus `SequenceExhausted(#[from] …)`, `IdentityMismatch { detail }` and `UnsupportedVersion { version, detail }`
- `SequenceManager::{allocate_sender_seq, increment_target_seq}` are `#[deprecated]`; removal queued for the next breaking release

## Tests

Workspace goes from 215 to **248 passing**; the engine integration suite from 11 to 35. New coverage includes both SequenceReset modes and the backward guard, the bounded/unbounded gap-fill replies, PossDup duplicate handling, too-low without `43=Y`, gapped Logon ack, `141=Y` reset, CompID mismatch, logout-ack timeout, drop-all-handles auto-logout, `from_admin` rejection, undecodable and 34-less frames, `with_initial_sequences` seeding, and an exhausted counter failing the handshake.

Two are worth singling out because they pin the crux of a fix rather than its surface: the resend test now sends an application message *after* the gap fill and asserts it goes out at 5, proving the fill consumed no sender number (it would be 6 otherwise); and the regression test above was checked in both directions.

## Verification

```
cargo test --workspace                                    # 248 passed, 0 failed
make doc                                                  # clean (clippy -W missing-docs)
```

No performance claim is made — the repo still has no benchmark harness.

### DIFF
--- a/doc/fix_operations.md
+++ b/doc/fix_operations.md
@@ -52,6 +52,50 @@ These messages are required for establishing and maintaining FIX sessions.
 4. Respond with Logon message
 5. Begin heartbeat monitoring
 
+**What IronFix implements today (initiator side)**
+
+`ironfix-engine::Initiator` validates the Logon acknowledgement in this order,
+and stops at the first failure:
+
+1. **MsgType.** A Logout or Reject is a rejected Logon; anything else is an
+   unexpected message.
+2. **Identity.** Inbound `SenderCompID` (49) must equal the configured
+   `target_comp_id` and inbound `TargetCompID` (56) the configured
+   `sender_comp_id`; when `sender_sub_id` / `target_sub_id` are configured,
+   inbound `TargetSubID` (57) and `SenderSubID` (50) are checked the same way.
+   A mismatch produces a session Reject with reason 9 (CompID problem) followed
+   by a Logout, and the handshake fails. This is checked before any callback
+   runs and before sequence state moves, so a cross-wired connection can never
+   establish a session.
+3. **`from_admin`.** A rejection sends a Logout and fails the handshake.
+4. **`HeartBtInt` (108).** The interval confirmed by the counterparty wins.
+   Note the counterparty therefore controls this value, and `108=0` — legal
+   FIX for "do not send heartbeats" — is not handled correctly today; see the
+   Phase 1 checklist entry for Heartbeat / Test Request.
+5. **`ResetSeqNumFlag` (141).** `141=Y` on the ack must arrive under
+   `MsgSeqNum = 1` — the reset and the number carrying it have to describe the
+   same stream — and a peer that sends any other number fails the handshake
+   rather than having one half of the contradiction guessed for it. It is
+   honored **before** MsgSeqNum is validated — otherwise the ack's
+   `34=1` reads as fatally too low against continuity-seeded counters. The
+   inbound counter is reset to 1. The outbound counter is set to 2, not 1.
+   That is an interpretive choice, not a derivation: it is exact when
+   `reset_on_logon` was set (the Logon on the wire really was message 1 of the
+   reset stream, and rewinding would re-emit a number the peer has seen), but
+   with continuity-seeded counters the Logon went out under its seeded number
+   and nothing numbered 1 was ever sent. QuickFIX/J's two halves disagree here
+   — its initiator would set 1, its acceptor 2 — and IronFix matches the
+   acceptor. The mismatch self-heals: the peer sees a gap at 2, requests a
+   resend of 1, and receives a GapFill that resynchronises it.
+6. **`MsgSeqNum` (34).** Too low fails the handshake; a gap completes the
+   handshake and immediately issues a ResendRequest.
+
+The same identity check (step 2) runs on every inbound frame once the session
+is established; there a mismatch produces the Reject and Logout and then tears
+the session down.
+
+Outbound `ResetSeqNumFlag` is driven by `SessionConfig::reset_on_logon`.
+
 ---
 
 ### Logout (MsgType = 5)
@@ -151,6 +195,40 @@ These messages are required for establishing and maintaining FIX sessions.
 3. Use SequenceReset-GapFill for admin messages
 4. Maintain original SendingTime with OrigSendingTime
 
+**What IronFix implements today**
+
+`ironfix-engine` has **no message store dependency**, so mandate items 1, 2 and
+4 above are **not implemented**: the engine cannot replay a single stored
+message. What it does is answer the whole requested range with one
+SequenceReset-GapFill (item 3, generalized to every message type):
+
+- `BeginSeqNo` (7) absent or unparseable → session Reject, reason 1 (required
+  tag missing), `RefTagID` = 7. There is deliberately **no** default value;
+  defaulting a missing BeginSeqNo to 1 silently answers a request the
+  counterparty never made.
+- `EndSeqNo` (16) absent or unparseable → session Reject, reason 1,
+  `RefTagID` = 16.
+- `BeginSeqNo` = 0, or at or beyond our next outbound sequence number → session
+  Reject, reason 5 (value incorrect), `RefTagID` = 7. We cannot resend what we
+  have not sent.
+- `EndSeqNo` below `BeginSeqNo` (and not 0) → session Reject, reason 5,
+  `RefTagID` = 16.
+- Otherwise the reply is `SequenceReset`-GapFill with `MsgSeqNum` = BeginSeqNo
+  and `NewSeqNo` = min(EndSeqNo + 1, next outbound sequence), or the next
+  outbound sequence when `EndSeqNo` = 0 (infinity). A bounded request therefore
+  never advances the counterparty past what it asked for.
+
+The GapFill carries `PossDupFlag` (43) = Y and `OrigSendingTime` (122). Absent
+a store there is no recorded original sending time, so 122 is stamped with the
+same value as `SendingTime` (52), which is the FIX handling for an unavailable
+OrigSendingTime. A GapFill's "original" messages are administrative filler
+rather than replayed traffic, so no information is lost by this.
+
+Consequence for consumers: **application messages are never replayed.** A
+counterparty that requests a resend of business traffic receives a gap fill,
+not the traffic. Real replay requires wiring a `MessageStore` into the engine,
+which is separate, tracked work.
+
 ---
 
 ### Reject (MsgType = 3)
@@ -190,6 +268,22 @@ These messages are required for establishing and maintaining FIX sessions.
 | 11 | Invalid MsgType |
 | 99 | Other |
 
+**Reason codes IronFix emits today**
+
+| Code | Emitted when |
+|---|---|
+| 1 | `SequenceReset` without `NewSeqNo` (36); `ResendRequest` without `BeginSeqNo` (7) or `EndSeqNo` (16) |
+| 5 | `SequenceReset` whose `NewSeqNo` would rewind or fails to advance; `ResendRequest` whose range is outside what we have sent |
+| 9 | Inbound `SenderCompID`/`TargetCompID` (and `SenderSubID`/`TargetSubID` when configured) do not match the session configuration |
+| any | Whatever an `Application::from_admin` / `from_app` implementation returns in its `RejectReason` |
+
+**Reason 10 (SendingTime accuracy) is not implemented.** `SendingTime` (52) is
+never compared against the local clock. Doing so requires a tolerance window,
+and a tolerance is a policy decision — it needs its own typed
+`SessionConfig` knob with a defensible default and a documented unit and range,
+not a guessed constant. Until that decision is made, a message with a wildly
+skewed `SendingTime` is accepted. This is a known, deliberate omission.
+
 ---
 
 ### Sequence Reset (MsgType = 4)
@@ -212,6 +306,37 @@ These messages are required for establishing and maintaining FIX sessions.
 **Server Behavior**:
 - **Gap Fill Mode**: Skip sequence numbers for admin messages during resend
 - **Reset Mode**: Force sequence number reset (use with caution)
+
+**What IronFix implements today**
+
+The two modes are **not** interchangeable, and `GapFillFlag` (123) is what
+selects between them. A Gap Fill is an ordinary sequenced message: it occupies
+its own `MsgSeqNum` and is validated like any other inbound message. **Reset
+mode is the only mode allowed to ignore `MsgSeqNum`.** Applying a gapped Gap
+Fill as though it were a Reset jumps the inbound expectation past messages that
+were never received and will now never be requested — a silent, permanent loss
+of (potentially) Execution Reports.
+
+**Not bounded:** `NewSeqNo` is accepted at any magnitude, so a peer can jump the
+inbound expectation arbitrarily far ahead — up to `u64::MAX`, after which the
+next message exhausts the counter and the session is torn down with a typed
+error. There is no principled ceiling in the specification to check against, and
+the failure mode is a clean teardown rather than silent corruption, so no
+arbitrary limit is invented here.
+
+`ironfix-engine` handles an inbound `SequenceReset` as follows:
+
+| Condition | Behavior |
+|---|---|
+| `from_admin` rejects the message | session Reject with the application's reason; `NewSeqNo` not applied |
+| `NewSeqNo` (36) absent or unparseable | session Reject, reason 1, `RefTagID` = 36; MsgSeqNum is **not** validated and **not** consumed, matching QuickFIX/J, which excludes SequenceReset from its increment-on-reject rule. For a GapFill this leaves the peer's next in-order message looking gapped, so expect a ResendRequest to follow |
+| `123=Y`, MsgSeqNum gapped | ResendRequest for the missing range; `NewSeqNo` **not** applied |
+| `123=Y`, MsgSeqNum too low with `PossDupFlag` = Y | dropped as an already-applied duplicate |
+| `123=Y`, MsgSeqNum too low without `PossDupFlag` | Logout and disconnect, as for any other too-low message |
+| `123=Y`, MsgSeqNum in sequence, `NewSeqNo` ≤ MsgSeqNum | session Reject, reason 5, `RefTagID` = 36; the fill message itself is consumed so the session does not deadlock on that number |
+| `123=Y`, MsgSeqNum in sequence, `NewSeqNo` > MsgSeqNum | applied: inbound expectation becomes `NewSeqNo` |
+| `123=N` or 123 absent (Reset mode), `NewSeqNo` < expected | session Reject, reason 5, `RefTagID` = 36; not applied |
+| `123=N` or 123 absent (Reset mode), `NewSeqNo` ≥ expected | applied regardless of MsgSeqNum. An outstanding ResendRequest is only cleared when the reset actually advances the expectation; a reset landing on the number already expected changes nothing, and clearing for it would let a peer replay it to trigger a fresh ResendRequest every round |
 
 ---
 
@@ -705,10 +830,20 @@ Rationale for the two policies:
 
 ### Phase 1: Core Session Layer (Required)
 - [x] Logon / Logout
-- [x] Heartbeat / Test Request
+- [ ] Heartbeat / Test Request — **partial**: the interval is negotiated and
+  heartbeats, TestRequests and the silent-peer timeout all work at the
+  configured interval, but two cases in `ironfix-session::HeartbeatManager` are
+  wrong. `HeartBtInt = 0` (legal FIX, meaning "do not send heartbeats") is
+  taken literally as a zero-length interval, so the session floods the peer and
+  then times itself out; and a pending TestRequest is cleared only by a
+  Heartbeat echoing the matching `TestReqID`, so a peer that answers without
+  tag 112 — or whose Heartbeat is reordered behind application traffic — is
+  disconnected despite being demonstrably alive.
 - [x] Reject
 - [x] Sequence Reset
-- [x] Resend Request
+- [ ] Resend Request — **partial**: inbound requests are validated and answered
+  with a correctly bounded SequenceReset-GapFill, but the engine has no message
+  store, so no message is ever actually replayed. See "Resend Request" above.
 
 ### Phase 2: Order Entry (High Priority)
 - [ ] New Order Single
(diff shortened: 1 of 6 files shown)


ironfix-engine/src/initiator.rs
```diff
@@ -574,8 +729,314 @@ impl<A: Application> Reactor<A> {
...
+    }
+
+    async fn on_sequence_reset(
+        &mut self,
+        framed: &mut FixFramed,
+        phase: Phase,
+        raw: &RawMessage<'_>,
+        seq: u64,
+    ) -> Result<Phase, SessionClosed> {
+        let gap_fill = raw.get_field_str(123) == Some("Y");
```
joaquinbejar: 🔴 **Critical**: every value other than `123=Y`, including malformed `123=X` or an empty value, is classified as Reset mode. Reset mode bypasses MsgSeqNum validation, so a hostile frame such as `34=100,123=X,36=101` while expecting 2 advances the target to 101 and permanently skips 2 through 100. Match `Y`, `N`, and absence explicitly, reject every other value, and add a malformed-123 reactor test.

ironfix-engine/src/initiator.rs
```diff
@@ -574,8 +729,314 @@ impl<A: Application> Reactor<A> {
...
+
+        if let Err(reason) = self.application.from_admin(raw, &self.session_id).await {
```
joaquinbejar, line 954: 🔴 **Critical**: `from_admin` runs before GapFill sequence validation. If the application rejects a gapped fill, this sends a Reject instead of the required ResendRequest, and a too-low `43=Y` duplicate reaches the callback instead of being dropped. Validate the GapFill sequence result first, then invoke the callback only for an in-sequence fill, with rejecting-application tests for the gapped and duplicate cases.

ironfix-engine/src/initiator.rs
```diff
@@ -271,9 +318,44 @@ impl<A: Application + 'static> Initiator<A> {
                 }
             }
 
+            let ack_seq: u64 = raw.get_field_as(34)?;
+
+            // Identity before anything else: a cross-wired acceptor must not
+            // be allowed to establish a session or move sequence state.
+            if let Err(mismatch) = identity.validate(&raw) {
```
joaquinbejar, line 351: 🔴 **Critical**: the FIXT handshake validates identity but never validates the acknowledgement's `BeginString` or required `DefaultApplVerID(1137)`. A configured FIX 5.0 SP2 session can become Active after a FIX.4.4 acknowledgement or a FIXT acknowledgement with a missing or different 1137, leaving the peers on different transport or application versions. Validate both fields before callbacks or sequence mutation and reject mismatches.

ironfix-engine/src/initiator.rs
```diff
@@ -296,9 +378,41 @@ impl<A: Application + 'static> Initiator<A> {
...
+                }
+                tracing::info!(
+                    session = %session_id,
+                    "counterparty set ResetSeqNumFlag on the Logon ack: resetting sequence numbers"
+                );
+                runtime.sequences.set_target_seq(1);
+                // The Logon already on the wire is message 1 of the reset
+                // outbound stream; rewinding the sender counter to 1 would
+                // re-emit a MsgSeqNum the counterparty has already seen.
+                runtime.sequences.set_sender_seq(2);
```
joaquinbejar, line 432: 🔴 **Critical**: this always resets the sender to 2 even when the already-sent Logon used a continuity-seeded number such as 10 rather than `141=Y,34=1`. The first post-logon application message then goes out as 2 while the peer expects 1 and can be gap-filled away during recovery. Reject unsolicited resets or set the sender to 1 unless the outbound Logon itself was the reset message.

ironfix-engine/src/initiator.rs
```diff
@@ -574,8 +729,314 @@ impl<A: Application> Reactor<A> {
...
+    async fn on_resend_request(
+        &mut self,
+        framed: &mut FixFramed,
+        phase: Phase,
+        raw: &RawMessage<'_>,
+        seq: u64,
+    ) -> Result<Phase, SessionClosed> {
+        let ref_msg_type = MsgType::ResendRequest.as_str();
+
+        let Some(begin_seq) = raw.get_field_str(7).and_then(|s| s.parse::<u64>().ok()) else {
```
joaquinbejar, line 1055: 🟠 **Warning**: this collapses missing, empty, and malformed numeric fields into session Reject reason 1. FIX distinguishes a missing required tag as reason 1, an empty value as reason 4, an incorrect data format such as `7=abc` as reason 6, and a semantic range error as reason 5. Split presence from parsing for tags 7, 16, and 36 and cover each case once.

doc/fix_operations.md
```diff
@@ -151,6 +195,40 @@ These messages are required for establishing and maintaining FIX sessions.
...
+- `EndSeqNo` (16) absent or unparseable → session Reject, reason 1,
+  `RefTagID` = 16.
+- `BeginSeqNo` = 0, or at or beyond our next outbound sequence number → session
+  Reject, reason 5 (value incorrect), `RefTagID` = 7. We cannot resend what we
+  have not sent.
+- `EndSeqNo` below `BeginSeqNo` (and not 0) → session Reject, reason 5,
+  `RefTagID` = 16.
+- Otherwise the reply is `SequenceReset`-GapFill with `MsgSeqNum` = BeginSeqNo
+  and `NewSeqNo` = min(EndSeqNo + 1, next outbound sequence), or the next
+  outbound sequence when `EndSeqNo` = 0 (infinity). A bounded request therefore
```
joaquinbejar, line 220: 🟠 **Warning**: this documents `EndSeqNo=0` as infinity for every advertised version, but FIX 4.0 and 4.1 use 999999; zero became the infinity sentinel in FIX 4.2. Make the outbound ResendRequest sentinel version-aware and add coverage for both version families.

doc/fix_operations.md
```diff
@@ -213,6 +307,37 @@ These messages are required for establishing and maintaining FIX sessions.
...
+error. There is no principled ceiling in the specification to check against, and
+the failure mode is a clean teardown rather than silent corruption, so no
+arbitrary limit is invented here.
+
+`ironfix-engine` handles an inbound `SequenceReset` as follows:
+
+| Condition | Behavior |
+|---|---|
+| `from_admin` rejects the message | session Reject with the application's reason; `NewSeqNo` not applied |
+| `NewSeqNo` (36) absent or unparseable | session Reject, reason 1, `RefTagID` = 36; MsgSeqNum is **not** validated and **not** consumed, matching QuickFIX/J, which excludes SequenceReset from its increment-on-reject rule. For a GapFill this leaves the peer's next in-order message looking gapped, so expect a ResendRequest to follow |
```
joaquinbejar: 🟠 **Warning**: this says an in-sequence GapFill with missing or malformed tag 36 is not consumed, but `on_sequence_reset` calls `consume_if_in_sequence` before rejecting it. Choose the intended protocol behavior, make the implementation and documentation agree, and assert the resulting target sequence in the malformed-NewSeqNo test.

doc/fix_operations.md
```diff
@@ -151,6 +195,40 @@ These messages are required for establishing and maintaining FIX sessions.
 3. Use SequenceReset-GapFill for admin messages
 4. Maintain original SendingTime with OrigSendingTime
 
+**What IronFix implements today**
+
+`ironfix-engine` has **no message store dependency**, so mandate items 1, 2 and
```
joaquinbejar: 🟠 **Warning**: `ironfix-engine` does have an `ironfix-store` dependency in its Cargo manifest; the actual gap is that the engine does not integrate or use a store for replay. Reword this to preserve the documented crate DAG and accurately describe the missing behavior.

joaquinbejar (commented):
The recovery work is substantial and the focused engine and session tests pass, but malformed GapFill handling, callback ordering, FIXT acknowledgement validation, reset sequencing, and version-specific recovery rules still expose live-session desynchronization paths. GitHub does not allow requesting changes on this self-authored PR, so this COMMENT is intended as REQUEST_CHANGES.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed in 3943f6b — tag 123 is matched explicitly now: Y is GapFill, N or absent is Reset, and any other present value is a session Reject reason 6 (incorrect data format, RefTagID 123) rather than being silently coerced to Reset. Agreed the malformed value gained no extra power, but guessing a mode from an uninterpretable field was the wrong default.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed in 3943f6b — the GapFill's MsgSeqNum is classified (gap / too-low duplicate / in-sequence) before from_admin now, like every other inbound message. A gapped fill emits the required ResendRequest even if the application rejects it, and a too-low 43=Y duplicate is dropped without the callback ever seeing it; only an in-sequence fill or a Reset reaches from_admin. The rejected-in-sequence fill still consumes its number, so the earlier wedge stays fixed.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed the BeginString half in 3943f6b — the ack's tag 8 is compared to the configured transport BeginString (FIXT.1.1 for a 5.0 session) before any callback or sequence mutation, and a mismatch aborts the handshake with EngineError::BeginStringMismatch instead of taking a FIX.4.4 ack Active on a FIXT session. The DefaultApplVerID(1137) half stays stamped-not-enforced, per the documented decision in the version section.

ironfix-engine/src/initiator.rs
joaquinbejar: This one's actually intended and documented — fix_operations.md Logon step 5 covers set_sender_seq(2) as a self-healing choice: the peer sees a gap at 2, requests a resend of 1, gets a GapFill for 1, and the app message at 34=2 is still processed, not lost. It names the continuity-seeded scenario explicitly and matches QuickFIX/J's acceptor half, so I'm leaving it.

ironfix-engine/src/initiator.rs
joaquinbejar: This conforms to the binding spec as written — fix_operations.md mandates reason 1 for BeginSeqNo(7)/EndSeqNo(16)/NewSeqNo(36) absent or unparseable. The finer 1/4/6 split is more FIX-conformant in general and a fair future proposal, but adopting it is a spec+code change rather than a bug fix; note reason 5 is already emitted for the range cases (begin_seq==0/out-of-range, rewinding NewSeqNo).

doc/fix_operations.md
joaquinbejar: Confirmed — recorded in 3943f6b as a known limitation. We emit 16=0 for every configured version, which is the >=4.2 infinity sentinel, so an open-ended resend on a 4.0/4.1 session isn't framed the way those versions expect (they use 999999). I added the version caveat to the doc; making the outbound sentinel BeginString-aware is small, isolated follow-up work, low real-world impact given the legacy versions.

doc/fix_operations.md
joaquinbejar: Fixed the doc in 3943f6b — the SequenceReset row now states that an in-sequence GapFill with an unusable NewSeqNo IS consumed (it occupies that number even though NewSeqNo is unusable), which is the behavior that avoids the ResendRequest deadlock the doc itself flags; Reset mode still consumes nothing. So the spec now matches the implemented wire behavior.

doc/fix_operations.md
joaquinbejar: Fixed in 3943f6b — reworded to say the engine doesn't use the store for replay, rather than that it has no store dependency (it does declare ironfix-store in Cargo.toml). Keeps the documented crate DAG accurate about the real as-built gap.

PR #30, opened 2026-07-21
fix(fast): correct the stop-bit primitives and bound the presence map

### LINKED ISSUE #6: ironfix-fast: primitive codec corrupts wire values — sign-carry byte never emitted, unchecked overflows panic or silently wrap, empty-string encoding broken
## Overview

The FAST primitive codec is wrong at its core: encode_int silently corrupts every signed value outside -64..64 (a market-data price/quantity round-trips to a DIFFERENT number), decode_bytes reaches a slice-index panic (process abort under panic=abort) on an attacker-controlled length, decode_int silently wraps on long inputs, and the ASCII codec violates the FAST 1.1 empty-string/NUL encodings. No error-path or round-trip tests exist anywhere in the crate — which is exactly why none of this was caught.

**Priority: P0** — from the full-workspace audit of 2026-07-21 (multi-agent adversarially-verified review, cross-checked against `doc/code_audit_2026-07-21.md`; findings marked `[source: codex-audit]` originate there).

## Context

Crates / areas touched: `ironfix-fast`

### 1. encode_int emits wire-corrupt encodings — sign-carry byte never emitted (finding 27, CRITICAL)
`ironfix-fast/src/encoder.rs:103`

Unconditional `if v == 0 && !negative { break; }` / `if v == -1 && negative { break; }` at lines 103-108 fire before the sign-conflict check at 97-99, so the extra sign-carry byte (0x00/0x7F) is never pushed. Empirically verified in release: `encode_int(64)` → `[0xC0]` → decodes as **-64**; 100→-28, 127→-1, 8192→-8192, -65→+63, `i64::MAX`→-1, `i64::MIN`→0. Every value whose minimal top payload byte conflicts with the sign round-trips to a WRONG value — silent price/quantity corruption; only -64..64 is correct.

**Fix:** delete the unconditional breaks; loop until v is 0/-1, then push 0x00/0x7F when the MSB's bit 6 conflicts with the sign. Exhaustive round-trip tests at every 7-bit boundary plus i64::MIN/MAX.

### 2. decode_bytes: usize overflow on attacker-controlled length → slice-index panic (finding 28)
`ironfix-fast/src/decoder.rs:171`

`if *offset + length > data.len()` overflows for a u64::MAX length prefix: release wraps and the bounds check falsely passes, then `data[*offset..*offset + length]` panics (reproduced: "slice index starts at 10 but ends at 9" at decoder.rs:175); debug panics on the add; `as usize` truncates on 32-bit targets. One malformed FAST packet aborts the process under `panic = "abort"`.

**Fix:** subtraction-based check compared in u64 before any usize cast; test feeding a stop-bit u64::MAX length asserting `Err(UnexpectedEof)`.

### 3. decode_int has no overflow check — silently wraps instead of IntegerOverflow (finding 29)
`ironfix-fast/src/decoder.rs:113`

`result = (result << 7) | (byte & 0x7F) as i64;` with no headroom check (unlike decode_uint at line 69): 10+ continuation bytes yield Ok(garbage) — probe: 12-byte stream decodes Ok(-1). Silently wrong signed values (price deltas, mantissas) on a live feed.

**Fix:** headroom checks against `i64::MAX >> 7` / `i64::MIN >> 7` before each shift, still accepting valid 10-byte MIN/MAX encodings; boundary + over-long tests.

### 4. ASCII codec breaks FAST empty-string/NUL encodings (finding 31)
`ironfix-fast/src/decoder.rs:137`

FAST 1.1: `[0x80]` is the EMPTY string and `[0x00,0x80]` the one-char NUL string. decode_ascii has no special case (0x80 → "\0"); encode_ascii mirrors it ("\0" → [0x80], colliding with empty). Probe: `encode_ascii("")` → `[0x80]` → `Ok("\0")`. Conformant feeds' empty strings are delivered wrong.

**Fix:** special-case `[0x80]` and `[0x00,0x80]` in both directions; reject other leading-0x00 (overlong) forms as InvalidString; round-trip tests for "", "\0", normal strings.

### 5. encode_nullable_uint(Some(u64::MAX)) overflows — silently encodes NULL in release (finding 32)
`ironfix-fast/src/encoder.rs:156`

`Some(v) => self.encode_uint(v + 1)`: debug panics; release wraps to 0 which encodes as 0x80 — the NULL representation, so `Some(u64::MAX)` silently becomes `None` on the wire (probe confirmed). `()` return gives no error path.

**Fix:** return `Result` with `checked_add` → `FastError::IntegerOverflow`; document the `0..=u64::MAX-1` domain; tests.

### 6. encode_ascii silently corrupts non-ASCII; decode_ascii's InvalidString mapping is dead (finding 33)
`ironfix-fast/src/encoder.rs:136`

`self.buffer.push(b & 0x7F)` masks multi-byte UTF-8 into different ASCII bytes — corrupt wire data, no error. decode_ascii can only produce bytes <= 0x7F so its from_utf8 error mapping can never fire, hiding the absence of real validation.

**Fix:** `encode_ascii -> Result`, reject `!value.is_ascii()` with InvalidString; route real validation (overlong forms) to InvalidString in decode.

### 7. No error-path or boundary tests anywhere in ironfix-fast; encode_int has zero tests (finding 35)
`ironfix-fast/src/encoder.rs:216`

No test exercises ANY FastError (no UnexpectedEof, IntegerOverflow, or oversized-length tests); encode_int — the function with the critical corruption bug — is untested and there is no encode_int↔decode_int round-trip, which would have caught the sign-carry bug immediately; decode_int tests cover only single-byte ±1.

**Fix:** round-trip sweep at every 7-bit boundary and i64::MIN/MAX; truncation tests per primitive; 10+-byte overflow tests; huge decode_bytes length asserting Err not panic; u64::MAX uint round-trip.

## Tasks

- [ ] encode_int emits wire-corrupt encodings — sign-carry byte never emitted (finding 27, CRITICAL)
- [ ] decode_bytes: usize overflow on attacker-controlled length → slice-index panic (finding 28)
- [ ] decode_int has no overflow check — silently wraps instead of IntegerOverflow (finding 29)
- [ ] ASCII codec breaks FAST empty-string/NUL encodings (finding 31)
- [ ] encode_nullable_uint(Some(u64::MAX)) overflows — silently encodes NULL in release (finding 32)
- [ ] encode_ascii silently corrupts non-ASCII; decode_ascii's InvalidString mapping is dead (finding 33)
- [ ] No error-path or boundary tests anywhere in ironfix-fast; encode_int has zero tests (finding 35)

## Acceptance criteria

- [ ] encode_int↔decode_int round-trip test sweeps every 7-bit boundary value (±63, ±64, ±127, ±128, ±8191, ±8192, ...) plus i64::MIN and i64::MAX, and all pass with byte-exact minimal encodings
- [ ] decode_bytes with a stop-bit-encoded u64::MAX length returns Err(FastError::UnexpectedEof) — no panic in debug or release, on 64-bit and (by construction) 32-bit targets
- [ ] decode_int on an over-long (10+ byte) continuation stream returns Err(FastError::IntegerOverflow); valid 10-byte MIN/MAX encodings still decode
- [ ] encode_ascii("") emits [0x80] and decodes back to ""; "\0" round-trips via [0x00,0x80]; other leading-0x00 overlong forms return Err(InvalidString)
- [ ] encode_nullable_uint returns Result; Some(u64::MAX) returns Err(IntegerOverflow) and the 0..=u64::MAX-1 domain is documented
- [ ] encode_ascii returns Result and rejects non-ASCII input with InvalidString; no silent byte-masking remains
- [ ] Every FastError variant that is claimed reachable has at least one test constructing it via the public API
- [ ] cargo test passes in both debug (overflow-checks on) and release profiles

joaquinbejar:
## Summary

Fixes the six defects in the FAST primitive codec. Two are wire-level corruption reachable from any counterparty.

Closes #6.

## Stack

- **Base branch:** `stack/3-9-engine-conformance` (PR #25)
- **Stack:** #25 → **#6 (this PR)** → the remaining open issues
- Review against the base branch; merge bottom-up.

## What was wrong

**1. The signed encoder never emitted the sign-carry byte.** The unconditional loop-exit fired before the sign-conflict check, so every value whose most significant payload byte conflicted with the sign round-tripped to a **different number**: `64 → -64`, `100 → -28`, `127 → -1`, `8192 → -8192`, `-65 → +63`, `i64::MAX → -1`, `i64::MIN → 0`. Only `-64..64` survived. For a price or quantity that is silent corruption on the wire.

Verified fixed against the exact values above plus a sweep of 633 boundary values (every 7-bit boundary ±2, both signs, through `i64::MIN`/`MAX`) — zero failures. The encoder now shares its width constants with the decoder so the two cannot drift.

**2. `decode_bytes` panicked on an attacker-controlled length.** `offset + length > data.len()` overflowed for a `u64::MAX` length prefix: release wrapped, the bounds check passed, and the slice index panicked — under `panic = "abort"` that is a remote process kill. It now computes remaining space with checked arithmetic, narrows the declared length in the direction that cannot truncate, and validates before slicing. No allocation is sized from the declared length.

**3. `decode_int` wrapped silently** on an over-long continuation run instead of reporting overflow, while a legitimate ten-byte `i64::MIN` / `u64::MAX` still decodes — over-rejection would refuse valid market data.

**4. FAST 1.1 string forms were wrong.** The empty string now encodes as a lone stop byte, a NUL string round-trips, other overlong leading-zero forms are rejected, and `encode_ascii` returns a typed error instead of silently masking non-ASCII bytes.

**5. Nullable integer encoding overflowed** on the value at the top of its bias; it is now fallible with a documented domain.

**6. Presence maps were unbounded and lied when exhausted.** No ceiling on encoded size, and an exhausted map answered "field absent" — so a truncated map read as a message whose remaining fields were all missing. The size is now capped (64 bytes ≈ 448 fields) and exhaustion is an explicit error.

Workspace is already at 0.4.0. `FastError` gains `#[non_exhaustive]` and new variants; `encode_ascii` and `encode_nullable_uint` become fallible. `ironfix-fast` has no in-workspace consumer except `ironfix-example`, whose two FAST servers are updated for the fallible calls.

## Tests

Workspace goes to **301 passing**, in **both debug and release** — the distinction matters here, since two of these defects only manifest under one profile or the other.

## Verification

```
cargo test --workspace            # 301 passed
cargo test --workspace --release  # 301 passed
cargo clippy --all-targets --all-features -- -D warnings
cargo fmt --all --check
cargo build --release --examples
make doc
```

### DIFF
--- a/ironfix-fast/src/decoder.rs
+++ b/ironfix-fast/src/decoder.rs
@@ -8,11 +8,57 @@
 //!
 //! This module provides decoding of FAST-encoded messages using stop-bit
 //! encoding and presence maps.
+//!
+//! Every function here is an untrusted-input parser: the bytes arrive from a
+//! counterparty over a socket. Malformed input maps to a typed [`FastError`]
+//! with a resource ceiling — never a panic, never an allocation sized from a
+//! declared length that has not been validated against the bytes actually
+//! available.
 
 use crate::error::FastError;
 use crate::operators::DictionaryValue;
 use crate::pmap::PresenceMap;
 use std::collections::HashMap;
+
+/// Stop bit: set on the final byte of a stop-bit encoded value.
+pub(crate) const STOP_BIT: u8 = 0x80;
+
+/// Mask selecting the seven payload bits of a stop-bit encoded byte.
+pub(crate) const PAYLOAD_MASK: u8 = 0x7F;
+
+/// Number of payload bits carried by each stop-bit encoded byte.
+pub(crate) const PAYLOAD_BITS: usize = 7;
+
+/// Bit 6 of the most significant payload byte: the sign of a signed integer.
+pub(crate) const SIGN_BIT: u8 = 0x40;
+
+/// Maximum number of bytes a stop-bit encoded 64-bit integer may occupy.
+///
+/// A 64-bit value carries at most 64 significant bits at
+/// [`PAYLOAD_BITS`] bits per byte; a signed value needs nine payload bytes
+/// plus one sign-carry byte, so ten bytes is the longest legal encoding of
+/// either `i64::MIN` / `i64::MAX` or `u64::MAX`.
+///
+/// Anything longer is rejected as [`FastError::IntegerOverflow`]. This is the
+/// resource ceiling that stops a hostile stream from spinning the decoder over
+/// an unbounded run of continuation bytes whose value never trips the
+/// arithmetic check (a run of `0x00` or `0x7F` bytes accumulates no new
+/// magnitude).
+pub const MAX_INT_ENCODED_LEN: usize = 10;
+
+/// Reads one byte and advances `offset`.
+///
+/// # Errors
+/// Returns [`FastError::UnexpectedEof`] if `offset` is at or past the end of
+/// `data`.
+#[inline]
+pub(crate) fn read_byte(data: &[u8], offset: &mut usize) -> Result<u8, FastError> {
+    let byte = *data.get(*offset).ok_or(FastError::UnexpectedEof)?;
+    // `get` succeeded, so `*offset < data.len() <= isize::MAX`; the increment
+    // cannot overflow a `usize`.
+    *offset += 1;
+    Ok(byte)
+}
 
 /// FAST protocol decoder.
 #[derive(Debug)]
@@ -45,6 +91,9 @@ impl FastDecoder {
 
     /// Decodes an unsigned integer using stop-bit encoding.
     ///
+    /// Over-long encodings are tolerated up to [`MAX_INT_ENCODED_LEN`] bytes,
+    /// which is the longest legal encoding of `u64::MAX`.
+    ///
     /// # Arguments
     /// * `data` - The input bytes
     /// * `offset` - Current position (will be updated)
@@ -53,27 +102,33 @@ impl FastDecoder {
     /// The decoded unsigned integer.
     ///
     /// # Errors
-    /// Returns `FastError::UnexpectedEof` if data is incomplete.
+    /// Returns [`FastError::UnexpectedEof`] if the stop bit is never reached
+    /// before the end of `data`, or [`FastError::IntegerOverflow`] if the
+    /// encoding is longer than [`MAX_INT_ENCODED_LEN`] bytes or denotes a
+    /// value that does not fit in a `u64`.
     pub fn decode_uint(data: &[u8], offset: &mut usize) -> Result<u64, FastError> {
+        /// Value of one payload byte position.
+        const RADIX: u64 = 1 << PAYLOAD_BITS;
+
         let mut result: u64 = 0;
+        let mut consumed: usize = 0;
 
         loop {
-            if *offset >= data.len() {
-                return Err(FastError::UnexpectedEof);
-            }
-
-            let byte = data[*offset];
-            *offset += 1;
-
-            // Check for overflow
-            if result > (u64::MAX >> 7) {
+            if consumed == MAX_INT_ENCODED_LEN {
                 return Err(FastError::IntegerOverflow);
             }
+            let byte = read_byte(data, offset)?;
+            consumed += 1;
 
-            result = (result << 7) | (byte & 0x7F) as u64;
+            // `checked_mul` rejects any value that would lose significant bits.
+            // The product's low seven bits are zero, so the `|` below cannot
+            // carry and is exact.
+            result = result
+                .checked_mul(RADIX)
+                .ok_or(FastError::IntegerOverflow)?
+                | u64::from(byte & PAYLOAD_MASK);
 
-            // Check stop bit
-            if byte & 0x80 != 0 {
+            if byte & STOP_BIT != 0 {
                 break;
             }
         }
@@ -83,6 +138,10 @@ impl FastDecoder {
 
     /// Decodes a signed integer using stop-bit encoding.
     ///
+    /// The sign is taken from bit 6 of the first byte and the value is
+    /// accumulated in two's complement, so a leading sign-carry byte (`0x00`
+    /// for a positive value, `0x7F` for a negative one) decodes transparently.
+    ///
     /// # Arguments
     /// * `data` - The input bytes
     /// * `offset` - Current position (will be updated)
@@ -91,28 +150,38 @@ impl FastDecoder {
     /// The decoded signed integer.
     ///
     /// # Errors
-    /// Returns `FastError::UnexpectedEof` if data is incomplete.
+    /// Returns [`FastError::UnexpectedEof`] if the stop bit is never reached
+    /// before the end of `data`, or [`FastError::IntegerOverflow`] if the
+    /// encoding is longer than [`MAX_INT_ENCODED_LEN`] bytes or denotes a
+    /// value that does not fit in an `i64`.
     pub fn decode_int(data: &[u8], offset: &mut usize) -> Result<i64, FastError> {
-        if *offset >= data.len() {
-            return Err(FastError::UnexpectedEof);
-        }
+        /// Value of one payload byte position.
+        const RADIX: i64 = 1 << PAYLOAD_BITS;
 
-        let first_byte = data[*offset];
-        let negative = (first_byte & 0x40) != 0;
+        let first_byte = *data.get(*offset).ok_or(FastError::UnexpectedEof)?;
+        let negative = (first_byte & SIGN_BIT) != 0;
 
+        // Sign extension: a negative value starts from all-ones so the
+        // accumulated two's complement stays negative.
         let mut result: i64 = if negative { -1 } else { 0 };
+        let mut consumed: usize = 0;
 
         loop {
-            if *offset >= data.len() {
-                return Err(FastError::UnexpectedEof);
+            if consumed == MAX_INT_ENCODED_LEN {
+                return Err(FastError::IntegerOverflow);
             }
+            let byte = read_byte(data, offset)?;
+            consumed += 1;
 
-            let byte = data[*offset];
-            *offset += 1;
+            // `checked_mul` rejects any value that would lose significant bits
+            // in either direction. The product is a multiple of `RADIX`, so its
+            // low seven bits are zero and the `|` below is exact.
+            result = result
+                .checked_mul(RADIX)
+                .ok_or(FastError::IntegerOverflow)?
+                | i64::from(byte & PAYLOAD_MASK);
 
-            result = (result << 7) | (byte & 0x7F) as i64;
-
-            if byte & 0x80 != 0 {
+            if byte & STOP_BIT != 0 {
                 break;
             }
         }
@@ -122,6 +191,10 @@ impl FastDecoder {
 
     /// Decodes an ASCII string using stop-bit encoding.
     ///
+    /// Follows the FAST 1.1 string encodings: a lone stop byte (`0x80`) is the
+    /// empty string and `0x00 0x80` is the one-character NUL string. Any other
+    /// encoding beginning with `0x00` is over-long and is rejected.
+    ///
     /// # Arguments
     /// * `data` - The input bytes
     /// * `offset` - Current position (will be updated)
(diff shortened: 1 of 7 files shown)


joaquinbejar (commented):
The arithmetic hardening tests pass in debug and release, but the FAST codec still conflicts with FAST 1.1 on presence-map suffixes and overlong forms, rejects a legal nullable boundary, retains an attacker-sized allocation, and incompletely inventories its public API changes. GitHub does not allow requesting changes on this self-authored PR, so this COMMENT is intended as REQUEST_CHANGES.

ironfix-fast/src/decoder.rs
```diff
@@ -45,6 +101,9 @@ impl FastDecoder {
 
     /// Decodes an unsigned integer using stop-bit encoding.
     ///
+    /// Over-long encodings are tolerated up to [`MAX_INT_ENCODED_LEN`] bytes,
```
joaquinbejar, line 104: 🔴 **Critical**: FAST 1.1 §10.6.1 makes an overlong integer reportable ERR R6, but this deliberately accepts redundant encodings such as `0x00 0x81` for 1. That also affects template identifiers. Reject redundant leading zero groups for unsigned values and redundant sign-extension groups for signed values, keep any block-size exception on a separate path, and add exact R6 tests.

ironfix-fast/src/decoder.rs
```diff
@@ -91,37 +160,76 @@ impl FastDecoder {
...
+    /// Decodes a nullable unsigned integer.
+    ///
+    /// FAST represents a nullable unsigned integer with a bias of one: a lone
+    /// stop byte is NULL, and any other value is the encoded value minus one.
+    /// This is the counterpart of
+    /// [`FastEncoder::encode_nullable_uint`](crate::FastEncoder::encode_nullable_uint);
+    /// the two must agree, so the bias lives here rather than in every caller,
+    /// which is where an off-by-one reappears.
+    ///
+    /// The representable domain is therefore `0..=u64::MAX - 1`.
```
joaquinbejar, line 211: 🔴 **Critical**: this domain excludes a valid FAST value. Section 10.6.1 requires non-negative nullable integers to be biased by one and explicitly permits one extra significant bit, so `u64::MAX` must round-trip through the 65-bit entity `0x02 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x80`. Handle that boundary without narrowing the biased value to `u64` and add the byte-exact round trip.

ironfix-fast/src/pmap.rs
```diff
@@ -50,55 +80,84 @@ impl PresenceMap {
...
     #[inline]
-    pub fn next_bit(&mut self) -> bool {
-        if self.position < self.bits.len() {
-            let bit = self.bits[self.position];
-            self.position += 1;
-            bit
-        } else {
-            false
-        }
+    pub fn next_bit(&mut self) -> Result<bool, FastError> {
```
joaquinbejar, line 139: 🔴 **Critical**: FAST 1.1 §10.5 defines a presence map as having an infinite zero suffix, so exhausting the transmitted bytes must continue to answer absent. A legal one-byte map used by a template with later absent fields currently fails at the eighth bit. Preserve the implicit `false` suffix and enforce ERR R8 separately where the template-required bit count is known.

ironfix-fast/src/pmap.rs
```diff
@@ -50,55 +80,84 @@ impl PresenceMap {
...
-            for i in (0..7).rev() {
-                bits.push((byte >> i) & 1 == 1);
+            // Extract the seven payload bits, most significant first.
+            for shift in (0..PAYLOAD_BITS).rev() {
+                bits.push((byte >> shift) & 1 == 1);
             }
 
-            // Check stop bit (high bit)
-            if byte & 0x80 != 0 {
+            if byte & STOP_BIT != 0 {
```
joaquinbejar, line 105: 🔴 **Critical**: decode accepts an overlong presence map such as `0x40 0x80`, while FAST 1.1 §10.5 requires ERR R7 when a multi-byte map ends with an all-zero group; `encode` can emit the same form when its final chunk is all false. Reject it on decode, trim it on encode while retaining one byte, and add byte-exact R7 tests.

ironfix-fast/src/decoder.rs
```diff
@@ -164,16 +297,31 @@ impl FastDecoder {
...
             return Err(FastError::UnexpectedEof);
         }
 
-        let bytes = data[*offset..*offset + length].to_vec();
-        *offset += length;
+        let end = offset.checked_add(length).ok_or(FastError::UnexpectedEof)?;
+        let bytes = data
+            .get(*offset..end)
+            .ok_or(FastError::UnexpectedEof)?
+            .to_vec();
```
joaquinbejar, line 323: 🔴 **Critical**: this still allocates the peer-declared field length with `to_vec()`. Checking the length against remaining input prevents an out-of-bounds read but does not prevent an attacker-sized allocation; a large complete field duplicates the payload and can exhaust memory. Enforce an explicit byte-vector ceiling before allocation or return a borrowed slice, with boundary regression tests.

ironfix-fast/src/decoder.rs
```diff
@@ -91,37 +160,76 @@ impl FastDecoder {
...
+    /// the two must agree, so the bias lives here rather than in every caller,
+    /// which is where an off-by-one reappears.
+    ///
+    /// The representable domain is therefore `0..=u64::MAX - 1`.
+    ///
+    /// # Errors
+    /// Returns [`FastError::UnexpectedEof`] if the input ends mid-value, or
+    /// [`FastError::IntegerOverflow`] if the encoded value exceeds the
+    /// representable domain.
+    pub fn decode_nullable_uint(data: &[u8], offset: &mut usize) -> Result<Option<u64>, FastError> {
```
joaquinbejar, line 217: 🟠 **Warning**: this is a new public API, but the PR description's semver inventory omits it and also misses the breaking `PresenceMap` return-type changes plus the added constants, re-exports, and `is_exhausted`. Since these crates are published, enumerate the complete public delta so the release impact is reviewable. The staged `0.4.0` line is appropriate relative to published `0.3.1`; the issue is the incomplete inventory.

ironfix-fast/src/decoder.rs
joaquinbejar: Leaving this — tolerating overlong integer encodings is deliberate and bounded (capped at MAX_INT_ENCODED_LEN=10 bytes, checked_mul catches real overflow), so a hostile run of continuation bytes can't cost more than that and gains no magnitude, and template IDs still resolve identically. FAST minimality is primarily an encoder obligation, and accepting overlong encodings is safe and improves interop; if strict conformance becomes a project goal it's a low-priority strictness tightening, not a correctness or security defect.

ironfix-fast/src/decoder.rs
joaquinbejar: Confirmed but low impact — tracked in #48 for a follow-up. A full-range nullable uInt64 would need to carry u64::MAX (encoded as the 65-bit 2^64), which decode_uint rejects as overflow, so the representable domain is 0..=u64::MAX-1. We're internally consistent though — our own encoder refuses Some(u64::MAX) too — and u64::MAX is never a real market value, so practical impact is tiny. The u128-and-special-case-the-boundary fix is noted on the issue.

ironfix-fast/src/pmap.rs
joaquinbejar: Confirmed as a real FAST-semantics point, but the fix belongs in the template-aware layer, not primitive next_bit here. PresenceMap already exposes bit() and is_exhausted() so the template layer can honor the infinite zero suffix; the actionable locus is the transfer path in PR #40 (operators.rs:173), where next_bit()? currently propagates the error instead of reading absent. Since a too-short pmap is legal in FAST, length-based truncation detection would itself be wrong, so I'd rather keep next_bit's contract and fix the reliance in the template decode.

ironfix-fast/src/pmap.rs
joaquinbejar: Leaving this — accepting a non-minimal pmap with trailing zero bytes is harmless, since those zeros are semantically identical to the implicit infinite zero suffix FAST defines, and the map is bounded by MAX_PMAP_BYTES=64 so there's no DoS. Decoder rejection of overlong pmaps isn't a clearly-mandated FAST requirement and would risk rejecting benign peer output; it's also in tension with honoring that zero suffix, which we do want.

ironfix-fast/src/decoder.rs
joaquinbejar: Resolved by PR #40 (directly above this one) — it changes decode_bytes to borrow from the input (returns &[u8]) and drops the to_vec entirely. Worth noting the allocation was already bounded by bytes actually delivered (guarded by `if length > remaining`), not sized from the unvalidated declared length, so the classic over-allocation was already avoided — it was at worst a transient copy of already-delivered bytes, which #40 removes.

ironfix-fast/src/decoder.rs
joaquinbejar: Acknowledged — the public delta (new decoder APIs, the PresenceMap return-type and is_exhausted changes, the new MAX_INT_ENCODED_LEN/MAX_PMAP_BYTES constants, the re-exports) is being enumerated in the PR description for the 0.4.0 semver record. Agreed staged 0.4.0 is the right call here.

joaquinbejar (commented):
The current head still leaves the earlier FAST conformance threads unresolved and also rejects a legal non-nullable ASCII value that begins with NUL. CI and focused debug, release, and clippy checks pass, but wire compatibility remains blocking. GitHub does not allow requesting changes on this self-authored PR, so this COMMENT is intended as REQUEST_CHANGES.

ironfix-fast/src/encoder.rs
```diff
@@ -45,100 +58,116 @@ impl FastEncoder {
...
+        match bytes {
+            [] => {
+                self.buffer.push(STOP_BIT);
+                return Ok(());
+            }
+            [0x00] => {
+                self.buffer.extend_from_slice(&[0x00, STOP_BIT]);
+                return Ok(());
             }
+            [0x00, ..] => return Err(FastError::InvalidString),
```
joaquinbejar, line 152: 🔴 **Critical**: FAST 1.1 §10.6.3 permits a non-nullable string beginning with NUL by adding a zero preamble, so `"\0A"` must encode as `00 00 C1`. This branch rejects that legal value, and `decoder.rs:255` rejects the corresponding wire form. Encode the required preamble, decode it distinctly from the single-NUL form `00 80`, and add byte-exact round-trip tests.

Issue #26, opened 2026-07-21
ironfix-core: MsgType::Custom allocates per message on the zero-allocation decode path


joaquinbejar:
## Overview

`MsgType::from_str` ends in `other => Self::Custom(other.to_string())`, so any tag 35 value the enum does not recognise heap-allocates **once per message** on the decode success path. A counterparty sending an unknown or vendor-specific MsgType therefore forces an allocation on every single message, on the path `CLAUDE.md` states allocates nothing.

Found by hot-path review during #7 (PR #23), which touched this expression — it replaced a `.unwrap()` there with an `Infallible` match — but deliberately left the allocation alone as out of scope.

## Context

- `ironfix-core/src/message.rs` — `MsgType::from_str`, the `Custom` variant
- `ironfix-tagvalue/src/decoder.rs` — the decode path that calls it, once per message
- The zero-allocation contract is stated in `CLAUDE.md` under "Key Decisions"

## Tasks

- [ ] Back `MsgType::Custom` with an `ArrayString` rather than a `String` — a FIX MsgType is at most two characters per the specification, so the variant can be bounded as well as allocation-free
- [ ] Follow the existing precedent: `CompId` is already an `ArrayString<32>` for the same reason
- [ ] Decide and document what happens to a MsgType longer than the bound — it must be an explicit typed error, never a silent truncation
- [ ] Confirm no remaining per-message allocation on the decode success path

## Acceptance criteria

- [ ] Decoding a message with an unrecognised MsgType performs no heap allocation
- [ ] An over-long MsgType value returns a typed error rather than truncating
- [ ] Round-trip tests for a custom MsgType at the bound and one byte over it
- [ ] Public-surface impact stated: changing `MsgType::Custom`'s payload type is a breaking change to a published enum
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [ ] `cargo fmt --all --check` clean
- [ ] `cargo test --workspace` clean
- [ ] `make doc` clean

PR #31, opened 2026-07-21
fix(core): harden the shared vocabulary against injection and silent loss

### LINKED ISSUE #16: ironfix-core: vocabulary hardening — CompId header injection, is_admin misclassification, unchecked FieldTag/Timestamp/OwnedMessage invariants, dead error variants, missing #[non_exhaustive], and accessor test gaps
## Overview

A sweep of ironfix-core correctness and API-hygiene fixes: CompId accepts SOH/'=' and injects header fields into every outbound message; is_admin() misclassifies XMLnonFIX against the vendored dictionary; FieldTag boundaries are off-by-one vs the spec; Timestamp epoch conversions silently wrap or zero; OwnedMessage indexes unvalidated offsets (abort); plus dead DecodeError variants, duplicated MsgType tables with an equality trap, leaked arrayvec error types, a falsely spec-attributed doc, missing #[non_exhaustive] on all 10 public error enums, and untested money-critical accessors.

## Context

Crates / areas touched: `ironfix-core`

### 1. CompId performs no charset validation — header field injection (finding 14)
`ironfix-core/src/types.rs:230` (230-232), `ironfix-engine/src/wire.rs:85`

`ArrayString::from(s).ok().map(Self)` — length is the only check; the engine writes CompIds unescaped into 49/56 on every outbound message via put_str→put_raw, so a hostile CompId injects arbitrary header fields or corrupts framing. Core is the right chokepoint.

**Fix:** validate printable ASCII excluding SOH and '=' in CompId::new/from_str with a typed error; tests for SOH, '=', non-ASCII rejection.

### 2. is_admin() misclassifies XMLnonFIX ('n') (finding 15)
`ironfix-core/src/message.rs:295` (295-306)

FIX44.xml:1008 marks 'n' msgcat='admin' but is_admin() matches only the seven classic session types; initiator.rs:669 branches on is_admin(), so inbound 'n' routes through the app path — hard-coded list disagrees with the dictionary.

**Fix:** add `| Self::XmlMessage` with a doc note; extend test_msg_type_is_admin.

### 3. FieldTag boundary off-by-one; tag 0 documented-but-unenforced (finding 16)
`ironfix-core/src/field.rs:53` (53-62)

FIX reserves 5000-9999 for user-defined fields, but `self.0 >= 1 && self.0 <= 5000` / `self.0 > 5000` puts 5000 on the wrong side; FieldTag::new documents 'must be > 0' with no check, so FieldTag(0) is neither standard nor user-defined.

**Fix:** boundary at 4999/5000; add try_new or is_valid; tests at 0, 4999, 5000.

### 4. Timestamp epoch conversions silently lossy/wrapping (finding 17)
`ironfix-core/src/types.rs:132` (also 122, 201)

`timestamp_nanos_opt().unwrap_or(0) as u64` — negative wraps huge, out-of-range silently 0 (forbidden silent default); from_millis multiplies unchecked (`millis * 1_000_000`); to_datetime casts back to i64 going negative for wrapped values. The From impl is the natural landing point for parsed SendingTime.

**Fix:** TryFrom with a typed error (or documented clamp), checked_mul in from_millis, document the representable range; tests for pre-1970 and large millis.

### 5. OwnedMessage::get_field unchecked indexing on constructor-unvalidated offsets; from_raw pointer subtraction (finding 21)
`ironfix-core/src/message.rs:558` (and 490)

OwnedMessage::new is pub with arbitrary (tag, Range) offsets; `&self.buffer[range.clone()]` panics (abort) on out-of-range — e.g. offsets from a corrupted store, the type's core use case. from_raw's `f.value.as_ptr() - raw.buffer.as_ptr()` underflows/wraps for FieldRefs not borrowing from buffer.

**Fix:** buffer.get(range) in get_field (signature already Option — non-breaking); debug_assert/skip out-of-range fields in from_raw; document the invariant; test out-of-range offsets.

### 6. FixField::encode infallible/Vec-based vs FixMessage::encode Result (finding 18)
`ironfix-core/src/field.rs:332` vs `message.rs:612`

`fn encode(value: &Self::Value, buf: &mut Vec<u8>);` — a field value that cannot be legally encoded (SOH in string, length bound exceeded — what EncodeError variants exist for) has no error path, forcing future implementors to panic or emit corrupt bytes. Zero implementors exist today, so the change is cheapest now.

**Fix:** change to Result<(), EncodeError> in the 0.4 breaking batch, before derive/codegen implement it.

### 7. MsgType: duplicated 60-arm matches, Custom equality trap (finding 19)
`ironfix-core/src/message.rs:152` (152-223 vs 225-291, Custom at 149)

FromStr and as_str duplicate a 60-entry mapping (drift silent since the fallback is Custom/Infallible; tests spot-check 4 variants). Custom("D") != NewOrderSingle under derived Eq/Hash despite identical wire form — HashMap/equality trap.

**Fix:** generate both matches from one table; exhaustive from_str(as_str())==v test; normalize Custom in a smart constructor or impl Eq/Hash over as_str().

### 8. Dead DecodeError variants (finding 20)
`ironfix-core/src/error.rs:67`

InvalidMsgType (unreachable by design — FromStr Infallible), GroupCountMismatch, MessageTooLarge are never constructed anywhere; dead semver-live vocabulary misleads implementors.

**Fix:** wire them up (e.g. a tagvalue max-size ceiling using MessageTooLarge) or #[deprecated] and remove at the next breaking release.

### 9. CompId::FromStr leaks arrayvec::CapacityError; Side::TryFrom Error = () (finding 22)
`ironfix-core/src/types.rs:269` (and 379)

`type Err = arrayvec::CapacityError<()>;` couples the core public API to arrayvec's semver; `type Error = ();` gives callers errors that cannot flow into FixError — in the crate whose job is owning the error vocabulary. Also blocks the CompId charset fix.

**Fix:** small typed CompIdError (TooLong/InvalidChar) and a Display-bearing error for Side::TryFrom; batch with the 0.4 breaking set.

### 10. CompId doc falsely attributes the 32-byte cap to the FIX spec (finding 23)
`ironfix-core/src/types.rs:215`

`/// Maximum length is 32 characters as per FIX specification.` — FIX defines CompIDs as unbounded String; 32 is an IronFix engineering bound (and it is bytes, not characters).

**Fix:** reword: 32 bytes, an IronFix inline-storage bound; the spec does not bound CompID length.

### 11. All 10 public error enums lack #[non_exhaustive] (finding 105)
`ironfix-core/src/error.rs:20` (workspace-wide)

`rg non_exhaustive ironfix-*/src` returns nothing — FixError/DecodeError/EncodeError/SessionError/StoreError/DictionaryError/ValidationError/FastError/CodecError/EngineError are all exhaustively matchable; Phases 2-5 will add variants and cargo-semver-checks will flag each as breaking. Marking them is itself a one-time break — batch it and record an ADR.

**Fix:** #[non_exhaustive] on all public error enums in the next breaking bump + doc/adr/0001 recording the decision.

### 12. FieldRef parse-failure paths zero tests (finding 24)
`ironfix-core/src/field.rs:335` (335-390)

The money-critical as_decimal accessor has no test anywhere (including rejection of b"1.2.3"); as_u64 on non-numeric, as_bool on non-Y/N, as_char on multi-byte are untested.

**Fix:** as_decimal valid+garbage, as_u64 b"abc"/b"-1" asserting InvalidFieldValue{tag}, as_bool b"y"/b"YES", as_char multi-byte/non-ASCII.

### 13. types.rs boundary cases untested (finding 25)
`ironfix-core/src/types.rs:386` (386-454)

test_comp_id uses 6 bytes, too-long at 33 — the exact-capacity boundary (32) is untested, precisely where ArrayString-backed types regress; no from_str/Side::TryFrom round-trip/format_micros tests.

**Fix:** 32-byte success, 33-byte parse error, exhaustive 16-variant Side round-trip, format_micros on a known epoch.

## Tasks

- [ ] CompId performs no charset validation — header field injection (finding 14)
- [ ] is_admin() misclassifies XMLnonFIX ('n') (finding 15)
- [ ] FieldTag boundary off-by-one; tag 0 documented-but-unenforced (finding 16)
- [ ] Timestamp epoch conversions silently lossy/wrapping (finding 17)
- [ ] OwnedMessage::get_field unchecked indexing on constructor-unvalidated offsets; from_raw pointer subtraction (finding 21)
- [ ] FixField::encode infallible/Vec-based vs FixMessage::encode Result (finding 18)
- [ ] MsgType: duplicated 60-arm matches, Custom equality trap (finding 19)
- [ ] Dead DecodeError variants (finding 20)
- [ ] CompId::FromStr leaks arrayvec::CapacityError; Side::TryFrom Error = () (finding 22)
- [ ] CompId doc falsely attributes the 32-byte cap to the FIX spec (finding 23)
- [ ] All 10 public error enums lack #[non_exhaustive] (finding 105)
- [ ] FieldRef parse-failure paths zero tests (finding 24)
- [ ] types.rs boundary cases untested (finding 25)

## Acceptance criteria

- [ ] CompId::new/from_str reject SOH, '=', and non-printable/non-ASCII bytes with a typed CompIdError; Side::TryFrom has a Display-bearing error; arrayvec::CapacityError no longer appears in any pub signature
- [ ] MsgType::is_admin() includes XmlMessage ('n') matching FIX44.xml, with an extended test
- [ ] FieldTag standard/user-defined boundary is 4999/5000; tag 0 is rejected by try_new/is_valid; tests at 0, 4999, 5000
- [ ] Timestamp conversions are TryFrom/checked (or documented clamps) — no unwrap_or(0), no unchecked multiply; pre-1970 and large-millis tests exist
- [ ] OwnedMessage::get_field uses buffer.get (no panic on corrupt offsets, tested); the from_raw borrow invariant is documented and debug_asserted
- [ ] FixField::encode returns Result<(), EncodeError> (batched in the 0.4 breaking set before any implementor lands)
- [ ] MsgType round-trip is table-generated with an exhaustive from_str(as_str())==v test; Custom normalization or wire-form equality is implemented and tested
- [ ] Dead DecodeError variants are each wired up or #[deprecated]; all 10 public error enums are #[non_exhaustive] with an ADR recording the batched break
- [ ] CompId doc no longer attributes the 32-byte bound to the FIX spec
- [ ] Tests added for as_decimal (valid + b"1.2.3"), as_u64/as_bool/as_char error paths, CompId at exactly 32/33 bytes, full Side round-trip, and format_micros

joaquinbejar:
## Summary

A sweep of `ironfix-core` correctness and API hygiene. Core is the root of the crate DAG, so every one of these defects is reachable from every crate above it.

Closes #16.

## Stack

- **Base branch:** `fix/6-fast-primitive-codec` (PR #30)
- **Stack:** #25 → #30 → **#16 (this PR)** → the remaining open issues
- Review against the base branch; merge bottom-up.

## What was wrong

**Header injection via `CompId`.** The constructor checked only length, and the engine writes CompIds unescaped into tags 49/56 on **every outbound message**. A hostile identifier containing SOH or `=` therefore injected arbitrary header fields or corrupted framing. Now rejected with a typed error at the constructor — core is the right chokepoint, since validating downstream would leave every other caller exposed.

**`is_admin()` disagreed with the dictionary.** `XMLnonFIX` ('n') is `msgcat='admin'` in the vendored `FIX44.xml`, but the hard-coded list omitted it, so inbound 'n' routed through the application path instead of the session path.

**`Timestamp` lost data silently.** An out-of-range instant became zero — a forbidden silent default, on the type that carries a parsed SendingTime — negative values wrapped through a `u64` cast, and `from_millis` multiplied unchecked. The conversions are fallible now, with the representable range documented.

**`FieldTag`** put 5000 on the wrong side of the user-defined boundary (FIX reserves 5000–9999) and documented "must be > 0" without enforcing it.

**`OwnedMessage::from_raw`** computed field offsets by subtracting raw pointers, which wraps for any `FieldRef` that does not borrow from the message buffer.

**`FixField::encode` was infallible** while `FixMessage::encode` returns `Result`, so a value that cannot be legally encoded — SOH inside a string, a length bound exceeded, exactly what `EncodeError` exists for — had nowhere to report it. There are no implementors yet, so it becomes fallible now, before `ironfix-derive` and `ironfix-codegen` build on it.

**`MsgType` had a silent-drift risk and an equality trap.** `FromStr` and `as_str` duplicated a sixty-arm mapping that could drift undetected, because the fallback is `Custom` and `FromStr` is `Infallible`. Both directions now come from one table. Separately, `Custom("D")` compared unequal to `NewOrderSingle` despite encoding identical bytes, so a `Custom` value missed its own entry in any `MsgType`-keyed map; equality and `Hash` now follow the on-the-wire form.

Dead `DecodeError` variants are removed.

Workspace is at 0.4.0, which covers all of it. Breaking: `CompId` construction is fallible; `Timestamp` epoch conversions are fallible; `FixField::encode` returns `Result`; `MsgType`'s `PartialEq`/`Hash` are hand-written rather than derived; dead `DecodeError` variants removed.

Fallout outside `ironfix-core` was mechanical and is limited to three call sites (`ironfix-derive`, `ironfix-engine/src/wire.rs`, and one engine test).

## Not in scope

`MsgType::Custom` still owns a `String`, so an unrecognised MsgType still allocates once per message. That is issue #26, deliberately kept separate so the two changes do not collide.

## Tests

Workspace goes from 304 to **343 passing**, including the previously untested money-critical accessors.

## Verification

```
cargo test --workspace                                    # 343 passed
```

### DIFF
--- a/ironfix-core/src/error.rs
+++ b/ironfix-core/src/error.rs
@@ -64,10 +64,6 @@ pub enum DecodeError {
     #[error("missing msg type field (tag 35)")]
     MissingMsgType,
 
-    /// Invalid MsgType value.
-    #[error("invalid msg type: {0}")]
-    InvalidMsgType(String),
-
     /// Checksum mismatch between calculated and declared values.
     #[error("checksum mismatch: calculated {calculated}, declared {declared}")]
     ChecksumMismatch {
@@ -97,29 +93,9 @@ pub enum DecodeError {
         reason: String,
     },
 
-    /// Repeating group count mismatch.
-    #[error("group count mismatch for tag {count_tag}: expected {expected}, found {actual}")]
-    GroupCountMismatch {
-        /// The tag containing the group count.
-        count_tag: u32,
-        /// Expected number of group entries.
-        expected: u32,
-        /// Actual number of group entries found.
-        actual: u32,
-    },
-
     /// Invalid UTF-8 in string field.
     #[error("invalid utf-8 in field: {0}")]
     InvalidUtf8(#[from] std::str::Utf8Error),
-
-    /// Message exceeds maximum allowed size.
-    #[error("message too large: {size} bytes exceeds maximum {max_size}")]
-    MessageTooLarge {
-        /// Actual message size in bytes.
-        size: usize,
-        /// Maximum allowed size in bytes.
-        max_size: usize,
-    },
 
     /// A field is not terminated by the SOH delimiter.
     ///
@@ -325,6 +301,120 @@ pub enum StoreError {
     /// I/O error in persistent store.
     #[error("store i/o error: {0}")]
     Io(String),
+}
+
+/// Rejection reasons for [`crate::types::CompId`] construction.
+///
+/// A CompID is written verbatim into SenderCompID (49) and TargetCompID (56)
+/// on every outbound message, so a value carrying SOH or `=` would inject
+/// header fields into the frame. Construction is the chokepoint that makes
+/// that unrepresentable.
+#[derive(Debug, Error, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum CompIdError {
+    /// The value does not fit in the fixed inline storage.
+    #[error("comp id is {len} bytes, exceeding the {max_len}-byte inline storage bound")]
+    TooLong {
+        /// Length of the offered value in bytes.
+        len: usize,
+        /// Maximum length in bytes, [`crate::types::COMP_ID_MAX_LEN`].
+        max_len: usize,
+    },
+
+    /// The value contains a byte outside printable ASCII, or the `=`
+    /// tag/value separator.
+    #[error(
+        "comp id contains illegal byte {byte:#04x} at offset {position}: \
+         only printable ASCII (0x20..=0x7e) except '=' is allowed"
+    )]
+    IllegalByte {
+        /// The offending byte.
+        byte: u8,
+        /// Zero-based offset of the offending byte within the value.
+        position: usize,
+    },
+}
+
+/// Rejection reasons for [`crate::types::Timestamp`] construction.
+///
+/// A `Timestamp` counts nanoseconds since the Unix epoch as an unsigned value
+/// bounded by [`crate::types::Timestamp::MAX_NANOS`], so instants before
+/// 1970-01-01 and after 2262-04-11 are not representable.
+#[derive(Debug, Error, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum TimestampError {
+    /// The nanosecond count exceeds the representable range.
+    #[error("{nanos} nanoseconds since the epoch exceeds the maximum {max_nanos}")]
+    NanosOutOfRange {
+        /// The offered nanosecond count.
+        nanos: u64,
+        /// The largest representable nanosecond count.
+        max_nanos: u64,
+    },
+
+    /// The millisecond count overflows when scaled to nanoseconds.
+    #[error("{millis} milliseconds since the epoch is not representable in nanoseconds")]
+    MillisOutOfRange {
+        /// The offered millisecond count.
+        millis: u64,
+    },
+
+    /// A calendar instant falls outside the representable range — before the
+    /// Unix epoch, or past the nanosecond ceiling.
+    #[error("instant at {seconds} seconds from the epoch is outside the representable range")]
+    InstantOutOfRange {
+        /// Whole seconds from the Unix epoch, negative before 1970.
+        seconds: i64,
+    },
+}
+
+/// A number that is not a legal FIX field tag.
+///
+/// FIX tags are positive integers; `0` is neither a standard nor a
+/// user-defined tag.
+#[derive(Debug, Error, Clone, Copy, PartialEq, Eq)]
+#[error("{tag} is not a legal FIX field tag: tags are positive integers starting at 1")]
+pub struct InvalidFieldTag {
+    tag: u32,
+}
+
+impl InvalidFieldTag {
+    /// Creates the error for an offending tag number.
+    #[inline]
+    #[must_use]
+    pub const fn new(tag: u32) -> Self {
+        Self { tag }
+    }
+
+    /// Returns the offending tag number.
+    #[inline]
+    #[must_use]
+    pub const fn tag(self) -> u32 {
+        self.tag
+    }
+}
+
+/// A byte that is not a legal Side (tag 54) code.
+#[derive(Debug, Error, Clone, Copy, PartialEq, Eq)]
+#[error("{value:#04x} is not a FIX Side (tag 54) code")]
+pub struct InvalidSide {
+    value: u8,
+}
+
+impl InvalidSide {
+    /// Creates the error for an offending byte.
+    #[inline]
+    #[must_use]
+    pub const fn new(value: u8) -> Self {
+        Self { value }
+    }
+
+    /// Returns the offending byte.
+    #[inline]
+    #[must_use]
+    pub const fn value(self) -> u8 {
+        self.value
+    }
 }
 
 #[cfg(test)]
--- a/ironfix-core/src/error.rs
+++ b/ironfix-core/src/error.rs
@@ -306,12 +306,19 @@ pub enum StoreError {
 /// Rejection reasons for [`crate::types::CompId`] construction.
 ///
 /// A CompID is written verbatim into SenderCompID (49) and TargetCompID (56)
-/// on every outbound message, so a value carrying SOH or `=` would inject
-/// header fields into the frame. Construction is the chokepoint that makes
-/// that unrepresentable.
+/// on every outbound message, so an empty value or one carrying SOH or another
+/// control byte would produce a malformed or injectable frame. Construction is
+/// the chokepoint that makes that unrepresentable.
 #[derive(Debug, Error, Clone, Copy, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum CompIdError {
+    /// The value is empty.
+    ///
+    /// An empty CompID would encode as `49=<SOH>` (or `56=<SOH>`), a field
+    /// with no value that FIX TagValue treats as malformed.
+    #[error("comp id is empty")]
+    Empty,
+
     /// The value does not fit in the fixed inline storage.
     #[error("comp id is {len} bytes, exceeding the {max_len}-byte inline storage bound")]
     TooLong {
@@ -321,11 +328,10 @@ pub enum CompIdError {
         max_len: usize,
     },
 
-    /// The value contains a byte outside printable ASCII, or the `=`
-    /// tag/value separator.
+    /// The value contains a byte outside printable ASCII (`0x20..=0x7e`).
     #[error(
         "comp id contains illegal byte {byte:#04x} at offset {position}: \
-         only printable ASCII (0x20..=0x7e) except '=' is allowed"
+         only printable ASCII (0x20..=0x7e) is allowed"
     )]
     IllegalByte {
         /// The offending byte.
--- a/ironfix-core/src/lib.rs
+++ b/ironfix-core/src/lib.rs
@@ -24,7 +24,10 @@ pub mod field;
 pub mod message;
 pub mod types;
 
-pub use error::{DecodeError, EncodeError, FixError, Result, SessionError, StoreError};
-pub use field::{FieldRef, FieldTag, FieldValue, FixField};
+pub use error::{
+    CompIdError, DecodeError, EncodeError, FixError, InvalidFieldTag, InvalidSide, Result,
+    SessionError, StoreError, TimestampError,
+};
+pub use field::{FieldRef, FieldTag, FieldValue, FixField, USER_DEFINED_TAG_MIN};
 pub use message::{FixMessage, MsgType, OwnedMessage, RawMessage};
-pub use types::{CompId, SeqNum, Side, Timestamp};
+pub use types::{COMP_ID_MAX_LEN, CompId, SeqNum, Side, Timestamp};
--- a/ironfix-core/src/message.rs
+++ b/ironfix-core/src/message.rs
@@ -20,133 +20,230 @@ use smallvec::SmallVec;
 use std::fmt;
 use std::ops::Range;
 
-/// Standard FIX message types.
+/// Declares [`MsgType`] and the three mappings that must agree about it.
 ///
-/// This enum covers the most common administrative and application messages.
-/// Custom or less common message types can be represented as `Custom(String)`.
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Default)]
-pub enum MsgType {
+/// The variant, its wire code, and its admin/app category are stated **once**
+/// per row; `as_str`, the `from_str` reverse lookup, and `is_admin` are all
+/// generated from that single table, so the three cannot drift. A row's
+/// category must match the `msgcat` the dictionary gives that MsgType — see
+/// [`MsgType::is_admin`].
+macro_rules! define_msg_types {
+    (
+        $(
+            $(#[$variant_meta:meta])*
+            $variant:ident = $code:literal, admin = $admin:literal;
+        )*
+    ) => {
+        /// Standard FIX message types.
+        ///
+        /// This enum covers the most common administrative and application
+        /// messages. Custom or less common message types are represented as
+        /// `Custom(String)`.
+        ///
+        /// # Equality follows the wire form
+        ///
+        /// [`PartialEq`] and [`Hash`] compare [`MsgType::as_str`], not the
+        /// variant: `MsgType::Custom("D".into()) == MsgType::NewOrderSingle`,
+        /// because both put `35=D` on the wire. Without that, a `Custom` value
+        /// reaching a `HashMap` keyed by `MsgType` would silently miss the
+        /// entry its own wire form should have hit. Prefer [`MsgType::new`],
+        /// which folds a known code into its named variant.
+        #[derive(Debug, Clone, Serialize, Deserialize, Default)]
+        pub enum MsgType {
+            $(
+                $(#[$variant_meta])*
+                $variant,
+            )*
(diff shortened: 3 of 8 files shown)


ironfix-core/src/types.rs
```diff
@@ -224,11 +343,31 @@ impl CompId {
...
+    /// bytes, or [`CompIdError::IllegalByte`] if it contains any byte outside
+    /// printable ASCII, or the `=` tag/value separator.
+    pub fn new(s: &str) -> Result<Self, CompIdError> {
+        if s.len() > COMP_ID_MAX_LEN {
+            return Err(CompIdError::TooLong {
+                len: s.len(),
+                max_len: COMP_ID_MAX_LEN,
+            });
+        }
+        for (position, &byte) in s.as_bytes().iter().enumerate() {
```
joaquinbejar, line 386: 🔴 **Critical**: this constructor is not the only construction path. The derived transparent `Deserialize` implementation writes the `ArrayString` directly and checks only capacity, so serialized `"SEND\u0001ER"` recreates an invalid `CompId`; the engine then writes that SOH verbatim into tags 49/56. Replace derived `Deserialize` with a validating implementation that calls `CompId::new`, and add hostile deserialization tests for control bytes and the length boundary.

ironfix-core/src/types.rs
```diff
@@ -224,11 +343,31 @@ impl CompId {
...
+            let printable = byte.is_ascii_graphic() || byte == b' ';
+            if !printable || byte == b'=' {
```
joaquinbejar: 🟠 **Warning**: `=` is legal inside a FIX String value. TagValue uses `=` between the numeric tag and its value, while SOH terminates the field, so `49=A=B<SOH>` is one field with value `A=B`, not an injected pair. Rejecting it breaks valid counterpart identifiers. Remove this restriction and correct the security rationale; retain the SOH/control-character check. See the FIX Trading Community TagValue specification: https://www.fixtrading.org/standards/tagvalue-online/

ironfix-core/src/types.rs
```diff
@@ -224,11 +343,31 @@ impl CompId {
...
-    /// `Some(CompId)` if the string fits within the maximum length, `None` otherwise.
-    #[must_use]
-    pub fn new(s: &str) -> Option<Self> {
-        ArrayString::from(s).ok().map(Self)
+    /// # Errors
+    /// Returns [`CompIdError::TooLong`] if `s` exceeds [`COMP_ID_MAX_LEN`]
+    /// bytes, or [`CompIdError::IllegalByte`] if it contains any byte outside
+    /// printable ASCII, or the `=` tag/value separator.
+    pub fn new(s: &str) -> Result<Self, CompIdError> {
+        if s.len() > COMP_ID_MAX_LEN {
```
joaquinbejar, line 380: 🟠 **Warning**: the length check and empty byte loop accept `CompId::new("")`, so the engine can emit malformed required fields as `49=<SOH>` or `56=<SOH>`. FIX TagValue explicitly treats an empty field value as malformed. Reject the empty string with a typed error and add constructor plus serialization-boundary tests.

ironfix-core/src/types.rs
```diff
@@ -92,45 +93,126 @@ impl fmt::Display for SeqNum {
...
+    ///
+    /// # Errors
+    /// Returns [`TimestampError::NanosOutOfRange`] if `nanos` exceeds
+    /// [`Timestamp::MAX_NANOS`].
     #[inline]
-    #[must_use]
-    pub const fn from_nanos(nanos: u64) -> Self {
-        Self {
-            nanos_since_epoch: nanos,
+    pub const fn from_nanos(nanos: u64) -> Result<Self, TimestampError> {
```
joaquinbejar, line 134: 🟠 **Warning**: derived `Deserialize` bypasses `from_nanos` and can construct `nanos_since_epoch > MAX_NANOS`. Formatting that value trips the new debug assertion in debug builds and silently substitutes `i64::MAX` in release, contradicting the total-by-construction guarantee. Deserialize through `Timestamp::from_nanos` and test `MAX_NANOS + 1`.

ironfix-core/src/message.rs
```diff
@@ -545,20 +527,37 @@ pub struct OwnedMessage {
...
+
+        let mut field_offsets = Vec::with_capacity(raw.fields.len());
+        for field in &raw.fields {
+            let offset = (field.value.as_ptr() as usize)
+                .checked_sub(buffer_start)
+                .and_then(|start| {
+                    let end = start.checked_add(field.value.len())?;
+                    (end <= buffer_len).then_some(start..end)
+                });
+            if let Some(range) = offset {
```
joaquinbejar, line 557: 🟠 **Warning**: dropping a field here turns a malformed `RawMessage` into a different apparently valid `OwnedMessage`. `RawMessage::new` is public and validates only its begin/body ranges, so a foreign `FieldRef` succeeds construction and then silently disappears during storage or cross-thread conversion. Validate every field slice against `buffer` in `RawMessage::new` or make `from_raw` fallible and return a typed error instead of losing fields.

ironfix-core/src/error.rs
```diff
@@ -327,6 +303,120 @@ pub enum StoreError {
...
 }
 
+/// Rejection reasons for [`crate::types::CompId`] construction.
+///
+/// A CompID is written verbatim into SenderCompID (49) and TargetCompID (56)
+/// on every outbound message, so a value carrying SOH or `=` would inject
+/// header fields into the frame. Construction is the chokepoint that makes
+/// that unrepresentable.
+#[derive(Debug, Error, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
```
joaquinbejar, line 313: 🟠 **Warning**: this PR says it closes #16, but that issue's acceptance criteria require every public error enum to become non-exhaustive and an ADR to record the batch. At this head `DictionaryError` and `ValidationError` remain exhaustive and there is no `doc/adr/`. Either complete that criterion and ADR or stop closing #16 and track the remaining work explicitly.

ironfix-core/src/lib.rs
```diff
@@ -24,7 +24,10 @@ pub mod field;
 pub mod message;
 pub mod types;
 
-pub use error::{DecodeError, EncodeError, FixError, Result, SessionError, StoreError};
-pub use field::{FieldRef, FieldTag, FieldValue, FixField};
+pub use error::{
```
joaquinbejar, line 27: 🟠 **Warning**: the PR description's public-API inventory is incomplete. It omits the breaking `Side::TryFrom<u8>::Error` change and FieldTag boundary behavior, plus these new error types, constructors, constants, re-exports, `MsgType::new`, and Timestamp APIs. Enumerate the complete public delta so downstream impact is reviewable. The staged `0.4.0` line is appropriate relative to published `0.3.1`; the issue is the missing inventory.

joaquinbejar (commented):
The focused core tests pass in debug and release, but the new CompId and Timestamp invariants are bypassable through Serde, malformed or valid CompID boundaries remain wrong, owned conversion can silently lose fields, and the issue/semver inventories are incomplete. GitHub does not allow requesting changes on this self-authored PR, so this COMMENT is intended as REQUEST_CHANGES.

joaquinbejar (commented):
The rebase onto PR #30 leaves the core tree unchanged, so the existing CompID and Timestamp deserialization bypasses and the other blocking review findings remain unresolved on this head. GitHub does not allow requesting changes on this self-authored PR, so this COMMENT is intended as REQUEST_CHANGES.

ironfix-core/src/types.rs
joaquinbejar: Good catch — fixed in 20de541. `CompId` now has a hand-written `Deserialize` that pulls a string and routes it through `CompId::new`, so a serialized value carrying SOH or any control byte errors during deserialization instead of rebuilding an invalid `CompId`. Added control-byte and length-boundary deserialization tests.

ironfix-core/src/types.rs
joaquinbejar: Fixed in 20de541 — I dropped the `=` check. `=` is legal inside a FIX value since a field splits on its first `=` only and SOH is what terminates it, so `49=A=B` is one valid field. The SOH/non-printable rejection stays, and I corrected the doc rationale that wrongly claimed `=` would open a new tag/value pair.

ironfix-core/src/types.rs
joaquinbejar: Fixed in 20de541 — `CompId::new` now rejects empty input with a new `CompIdError::Empty` variant (the enum is `#[non_exhaustive]`, so no semver break), matching what `MsgType::make_custom` already does. Added a constructor test.

ironfix-core/src/types.rs
joaquinbejar: Good catch — fixed in 20de541. `Timestamp` now deserializes the `u64` through `Timestamp::from_nanos`, so a value above `MAX_NANOS` errors instead of building a struct that later trips the `debug_assert` (or substitutes `i64::MAX` in release). Added a `MAX_NANOS + 1` test.

ironfix-core/src/message.rs
joaquinbejar: This one stands as is. The fields are `FieldRef`s holding real borrowed slices that are always valid to read regardless of provenance, so there's no bounds or memory-safety hole — that's why `RawMessage::new` only checks the begin/body ranges. `from_raw`'s pointer subtraction is checked and deliberately drops a field it can't re-base rather than wrapping (covered by the test at message.rs:924). Hitting the mismatch means calling the public `new()` with a foreign slice, which the type documents as unrepairable-by-construction.

ironfix-core/src/error.rs
joaquinbejar: All the core error enums are already `#[non_exhaustive]` on this head. The two the comment names — `DictionaryError` and `ValidationError` — live in `ironfix-dictionary` and get their `#[non_exhaustive]` in PR #39 upstack, so they're correctly still exhaustive here rather than a gap.

ironfix-core/src/lib.rs
joaquinbejar: Right, no code change here. The full public delta — `Side`'s `TryFrom<u8>` error type, the `FieldTag` boundary behavior, the new error types/variants, the constructors, `COMP_ID_MAX_LEN`, the re-exports, `MsgType::new`, and the `Timestamp` API — is being enumerated for the 0.4.0 semver record and routed to architect.

Issue #7, opened 2026-07-21
ironfix-tagvalue: decoder panics on the 2nd message from one Decoder, cannot parse FIX data fields (RawData/Signature/XmlData), and conflates malformed input with end-of-buffer


joaquinbejar:
## Overview

The tag=value decoder has three verified correctness bugs on untrusted input: absolute-vs-relative range confusion makes begin_string() panic (process abort) on the second message decoded from one Decoder; spec-defined Length/data field pairs containing SOH are either rejected or silently decoded with phantom injected fields; and structural garbage is indistinguishable from clean end-of-buffer, yielding Incomplete-forever or silent success. A build-mode-dependent checksum overflow panic completes the set.

**Priority: P0** — from the full-workspace audit of 2026-07-21 (multi-agent adversarially-verified review, cross-checked against `doc/code_audit_2026-07-21.md`; findings marked `[source: codex-audit]` originate there).

## Context

Crates / areas touched: `ironfix-tagvalue`, `ironfix-core`

### 1. RawMessage ranges are absolute but buffer is sliced — 2nd message panics (finding 0)
`ironfix-tagvalue/src/decoder.rs:168`, `ironfix-core/src/message.rs:376`

decoder.rs:83-85 computes begin_string/body ranges by pointer-diff against `self.input` (absolute offsets), but decoder.rs:168 passes the sub-slice `&self.input[start_offset..self.offset]` as the RawMessage buffer. For the second and subsequent messages, `begin_string()` indexes out of bounds and panics — reproduced: 'range start index 34 out of range for slice of length 32' at message.rs:376 — and `panic=abort` kills the process; `body_range()` is silently wrong. Additionally message.rs:376 `from_utf8(...).unwrap_or("")` maps invalid UTF-8 to "" — a forbidden silent default.

**Fix:** rebase both ranges by subtracting start_offset before constructing RawMessage; in core use `buffer.get()` and return a typed error instead of index + unwrap_or; document/debug_assert that RawMessage ranges are buffer-relative. Regression test: decode two consecutive frames with one Decoder.

### 2. Data fields (93/89, 95/96, 212/213, 354/355, 90/91) containing SOH cannot be parsed (finding 1)
`ironfix-tagvalue/src/decoder.rs:198`

[...]

PR #32, opened 2026-07-21
perf(core): make an unrecognised MsgType allocation-free

### LINKED ISSUE #26: ironfix-core: MsgType::Custom allocates per message on the zero-allocation decode path

joaquinbejar:
## Summary

Removes the last per-message heap allocation from the tag=value decode success path.

Closes #26.

## Stack

- **Base branch:** `stack/16-core-vocabulary` (PR #31)
- **Stack:** #25 → #30 → #31 → **#26 (this PR)** → the remaining open issues

## What was wrong

`MsgType::from_str` ended in `other => Self::Custom(other.to_string())`. Any tag 35 value the enum does not name therefore **heap-allocated once per message**, on the path `CLAUDE.md` states allocates nothing. A counterparty sending a vendor-specific MsgType forced an allocation on every single message — not an edge case, a steady state.

Found by hot-path review during #7 (PR #23), which touched this exact expression and deliberately left the allocation as out of scope.

## The fix

`Custom` now holds an `ArrayString` bounded by `MSG_TYPE_MAX_LEN`, so the value lives inside the enum. This follows the existing precedent — `CompId` is an `ArrayString<32>` for the same reason — and `arrayvec` was already a workspace dependency, so nothing new is pulled in. The bound leaves headroom over what the specification allows rather than sitting exactly on it.

**An over-long code is a typed error, never a truncation.** A silently truncated MsgType would route a frame to the wrong handler, which is worse than refusing it. That makes `FromStr` fallible where it was `Infallible` — and as a side effect `DecodeError::InvalidMsgType` becomes a variant the decoder actually constructs, rather than the dead surface it had been.

Equality and `Hash` still follow the on-the-wire form, so a `Custom` value keeps comparing equal to the named variant encoding the same bytes — the HashMap trap fixed in #31 stays fixed.

Breaking, covered by the in-flight 0.4.0: `MsgType::Custom`'s payload type changes, and `MsgType`'s `FromStr::Err` becomes a real error type.

## Tests

Workspace goes from 343 to **355 passing**, including the bound, the over-long rejection, and the round trip through the decoder.

## Verification

```
cargo test --workspace                                    # 355 passed
```

### DIFF
--- a/ironfix-core/src/message.rs
+++ b/ironfix-core/src/message.rs
@@ -43,46 +43,108 @@ pub const MSG_TYPE_MAX_LEN: usize = 8;
 
 /// Inline storage for a MsgType this enum does not name.
 ///
-/// Bounded by [`MSG_TYPE_MAX_LEN`], so it lives in the enum itself and never
-/// on the heap.
-pub type CustomMsgType = ArrayString<MSG_TYPE_MAX_LEN>;
+/// A validated newtype over an [`ArrayString`] bounded by [`MSG_TYPE_MAX_LEN`],
+/// so it lives in the enum itself and never on the heap. The inner buffer is
+/// **private** and reachable only through [`CustomMsgType::new`] (or the
+/// [`Deserialize`] impl, which routes the same constructor), so a
+/// [`MsgType::Custom`] can never be built around an empty code, an over-long
+/// code, or one carrying a byte that cannot appear in tag 35.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize)]
+pub struct CustomMsgType(ArrayString<MSG_TYPE_MAX_LEN>);
 
-/// Copies an unrecognised MsgType code into inline storage, validating it.
-///
-/// The bytes reach here straight off the wire — every decoded MsgType is built
-/// through [`MsgType::new`] — so this is where an untrusted tag 35 value is
-/// either bounded, non-empty, and safe to write back into a frame, or an
-/// error. `=` and SOH are rejected because a MsgType is echoed verbatim into
-/// tag 35 of an outbound message, where either byte would open a new tag/value
-/// pair or terminate the field early. The length bound is additionally
-/// enforced by [`CustomMsgType`] itself, so it holds even for a `Custom` built
-/// by hand.
-///
-/// # Errors
-/// Returns [`MsgTypeError::Empty`], [`MsgTypeError::TooLong`], or
-/// [`MsgTypeError::IllegalByte`].
-fn make_custom(code: &str) -> Result<CustomMsgType, MsgTypeError> {
-    if code.is_empty() {
-        return Err(MsgTypeError::Empty);
-    }
-    if code.len() > MSG_TYPE_MAX_LEN {
-        return Err(MsgTypeError::TooLong {
-            len: code.len(),
-            max_len: MSG_TYPE_MAX_LEN,
-        });
-    }
-    for (position, &byte) in code.as_bytes().iter().enumerate() {
-        if !byte.is_ascii_graphic() || byte == b'=' {
-            return Err(MsgTypeError::IllegalByte { byte, position });
+impl CustomMsgType {
+    /// Copies an unrecognised MsgType code into inline storage, validating it.
+    ///
+    /// This is the **only** way to obtain a `CustomMsgType`: the inner buffer
+    /// is private, so a [`MsgType::Custom`] cannot wrap an unvalidated code,
+    /// and the [`Deserialize`] impl routes the same check. The bytes typically
+    /// reach here straight off the wire — every decoded MsgType is built
+    /// through [`MsgType::new`], which sends an unnamed code here — so this is
+    /// where an untrusted tag 35 value is either bounded, non-empty, and safe
+    /// to write back into a frame, or an error.
+    ///
+    /// SOH and any non-printable or non-ASCII byte are rejected because a
+    /// MsgType is echoed verbatim into tag 35 of an outbound message, where SOH
+    /// would terminate the field early and a non-printable byte would corrupt
+    /// the frame. `=` and space are **not** rejected: both are legal inside a
+    /// FIX field value — only SOH ends a field, and only the *first* `=` splits
+    /// a tag from its value — so a bilaterally agreed code carrying either must
+    /// round-trip unchanged. The accepted set is therefore printable ASCII,
+    /// `0x20..=0x7e`. The length bound is additionally enforced by
+    /// [`MSG_TYPE_MAX_LEN`], so it holds even for a code built by hand.
+    ///
+    /// # Errors
+    /// Returns [`MsgTypeError::Empty`] for an empty code,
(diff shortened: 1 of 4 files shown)


ironfix-core/src/message.rs
```diff
@@ -12,303 +12,372 @@
...
+/// [`MsgType::Custom`] in inline [`ArrayString`] storage, so decoding a
+/// message whose MsgType this enum does not name allocates nothing. A value
+/// longer than this is [`MsgTypeError::TooLong`], never a truncation.
+pub const MSG_TYPE_MAX_LEN: usize = 8;
+
+/// Inline storage for a MsgType this enum does not name.
+///
+/// Bounded by [`MSG_TYPE_MAX_LEN`], so it lives in the enum itself and never
+/// on the heap.
+pub type CustomMsgType = ArrayString<MSG_TYPE_MAX_LEN>;
```
joaquinbejar: 🔴 **Critical**: `CustomMsgType` is a public `ArrayString` alias, so callers and derived `Deserialize` can bypass `MsgType::new` and construct `MsgType::Custom` with an empty value or embedded SOH. `MessageFactory` then writes it verbatim into tag 35, producing an invalid or injected frame. Replace this with an opaque validated newtype and deserialize through its checked constructor.

ironfix-core/src/message.rs
```diff
@@ -12,303 +12,372 @@
...
+/// by hand.
+///
+/// # Errors
+/// Returns [`MsgTypeError::Empty`], [`MsgTypeError::TooLong`], or
+/// [`MsgTypeError::IllegalByte`].
+fn make_custom(code: &str) -> Result<CustomMsgType, MsgTypeError> {
+    if code.is_empty() {
+        return Err(MsgTypeError::Empty);
+    }
+    if code.len() > MSG_TYPE_MAX_LEN {
```
joaquinbejar: 🟠 **Warning**: this eight-byte engineering ceiling rejects protocol-valid bilaterally defined MsgType values even though tag 35 is a FIX `STRING` with no such wire limit. The same validator rejects `=` and space even though SOH, not `=`, terminates a TagValue field. Preserve the valid bounded-by-frame STRING domain, using inline storage plus a bounded fallback if needed, and add fixtures above eight bytes and with legal punctuation.

ironfix-tagvalue/src/decoder.rs
```diff
@@ -190,12 +190,11 @@ impl<'a> Decoder<'a> {
...
-        // so the error arm is uninhabited rather than unwrapped.
-        let msg_type: MsgType = match msg_type_field.as_str()?.parse() {
-            Ok(parsed) => parsed,
-            Err(never) => match never {},
-        };
+        // An unrecognised code becomes `MsgType::Custom` in inline storage, so
+        // this allocates nothing; an empty, over-long, or byte-illegal value is
+        // `DecodeError::InvalidMsgType` rather than a truncated code that would
+        // route the frame to the wrong handler.
+        let msg_type: MsgType = msg_type_field.as_str()?.parse()?;
```
joaquinbejar, line 197: 🔴 **Critical**: this converts an invalid tag 35 into a decode failure, but `Initiator::on_frame` drops decode errors before reading tag 34, consuming `NextNumIn`, or sending the documented session Reject reason 11. An in-sequence bad MsgType followed by a valid next frame therefore leaves the session waiting on the dropped sequence number and can loop on resend. Preserve the structurally valid frame through sequencing and reject it with `371=35` and `373=11`, with an engine integration test that proves sequence advancement.

joaquinbejar (commented):
The inline storage goal is useful, but this head makes invalid MsgType a decoder failure that the engine silently drops before sequencing or Reject, exposes unchecked construction paths that can emit SOH in tag 35, and imposes an eight-byte limit outside the FIX String contract. The intended effective verdict is REQUEST_CHANGES. GitHub does not allow that on a self-authored PR, so this review is submitted as COMMENT.

ironfix-core/src/message.rs
joaquinbejar: Fixed in becf188 — `CustomMsgType` is now an opaque newtype with a private inner reachable only through the validating `make_custom`, and its `Deserialize` routes through that too, so an empty or illegal-byte code errors instead of being handed verbatim to tag 35. Flagged the new public shape to architect.

ironfix-core/src/message.rs
joaquinbejar: Keeping the 8-byte ceiling as is. It's a deliberate, documented resource ceiling that keeps `Custom` in inline `ArrayString` storage — an over-long code is a typed `MsgTypeError::TooLong`, never truncated, and a heap fallback would reintroduce the per-message allocation this PR exists to remove. Real MsgTypes (including `U`-prefixed bilaterals) fit easily. The one valid sub-point — relaxing the `=` rejection — is handled alongside the `CompId` `=` fix in #31 and in becf188's decoder change.

ironfix-tagvalue/src/decoder.rs
joaquinbejar: Partly addressed in becf188: the wire layer no longer hard-fails on a representable MsgType — `35=A=B` now decodes as a valid `Custom` value that round-trips, so an in-sequence frame carrying a `=` in tag 35 is no longer silently dropped. The remaining half — the engine's `on_frame` sequence-accounting a structurally-recoverable frame and sending a session `Reject` (`371=35`/`373=11`) instead of dropping it, which also affects bad-checksum frames — is a pre-existing engine/session issue tracked for the session layer, not this wire-layer PR.

PR #34, opened 2026-07-21
refactor(core): give the FIX version mapping a single home

### LINKED ISSUE #28: ironfix-core: hoist the FIX version mapping so engine and dictionary stop duplicating it
## Overview

The mapping from a configured FIX version to its wire representation — `BeginString` (tag 8), and `ApplVerID` for the 5.0 family — now exists in two places: `ironfix_dictionary::schema::Version` (`begin_string` / `appl_ver_id`) and a `pub(crate)` table added to `ironfix-engine` in #9 (PR #25).

The duplication is deliberate, not an oversight: `ironfix-engine` must not depend on `ironfix-dictionary` (a hard DAG invariant), so it cannot reuse the dictionary's answer. But the two tables can silently drift apart, and **no test can cross-check them** without taking the forbidden dependency.

## Context

- `ironfix-engine/src/wire.rs` — `WireVersion` / `wire_version`, with the debt already documented on the type
- `ironfix-dictionary/src/schema.rs` — `Version::begin_string`, `Version::appl_ver_id`
- `CLAUDE.md` — the DAG invariant, and "Cross-crate shared types live in the shared crate. Do not duplicate."

Both crates already depend on `ironfix-core`, which is the natural home.

## Tasks

- [ ] Add a FIX version type to `ironfix-core` owning the BeginString / ApplVerID mapping
- [ ] Have `ironfix-dictionary::schema::Version` and `ironfix-engine`'s wire layer both derive from it, deleting the duplicated tables
- [ ] Keep the engine's behaviour of refusing a version it cannot frame conformantly (unknown string, or bare `FIXT.1.1`, which names no application version for the required `DefaultApplVerID`)
- [ ] Remove the drift caveat from `wire.rs` once there is a single source

## Acceptance criteria

- [ ] Exactly one table in the workspace maps a FIX version to BeginString and ApplVerID
- [ ] A test asserts the mapping for every supported version, reachable from both consumers
- [ ] The crate DAG is unchanged: `ironfix-engine` still does not depend on `ironfix-dictionary`
- [ ] Public-surface impact stated — adding a type to `ironfix-core` is a semver event for all 11 crates
- [ ] `make doc` clean

joaquinbejar:
## Summary

Gives the FIX version → wire-form mapping a single home, removing a duplication that no test could detect drifting.

Closes #28.

## Stack

- **Base branch:** `stack/26-msgtype-alloc` (PR #32)
- **Stack:** #25 → #30 → #31 → #32 → **#28 (this PR)**

## What was wrong

The map from a configured FIX version to its wire form — `BeginString` (tag 8), and `ApplVerID` for the 5.0 family — existed in **two** places:

- `ironfix-dictionary/src/schema.rs`, in `Version::begin_string` / `Version::appl_ver_id`
- `ironfix-engine/src/wire.rs`, in a `pub(crate)` table added when FIXT.1.1 support landed in #25

The duplication was deliberate, not sloppiness: `ironfix-engine` must not depend on `ironfix-dictionary` (a hard DAG invariant), so it could not reuse the dictionary's answer. But it meant the two tables could drift apart silently, and **no test could cross-check them** without taking exactly the dependency the invariant forbids. Getting this wrong is not cosmetic — a wrong BeginString is rejected outright by a conforming acceptor.

## The fix

Both crates already depend on `ironfix-core`, so the mapping lives there and both derive from it. One table, one place, and the cross-check is now an ordinary test.

The engine's behaviour from #25 is preserved: a version it cannot frame conformantly — an unknown string, or bare `FIXT.1.1`, which names a transport version but no application version for the **required** `DefaultApplVerID` (1137) — is refused before dialling rather than guessed onto the wire.

## Crate DAG

Unchanged and verified: `ironfix-engine` still depends on core, session, store, transport and tagvalue, and **not** on `ironfix-dictionary`. `ironfix-core` still depends on nothing internal.

Adding a public type to `ironfix-core` is a semver event for all 11 crates; the workspace is already at 0.4.0, which covers it.

## Tests

Workspace goes to **367 passing**, including the mapping assertion that was previously impossible to write from a single place.

## Verification

```
cargo test --workspace                                    # 367 passed
```

### DIFF
--- a/ironfix-dictionary/src/schema.rs
+++ b/ironfix-dictionary/src/schema.rs
@@ -17,68 +17,19 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
 /// FIX protocol version.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
-pub enum Version {
-    /// FIX 4.0
-    Fix40,
-    /// FIX 4.1
-    Fix41,
-    /// FIX 4.2
-    Fix42,
-    /// FIX 4.3
-    Fix43,
-    /// FIX 4.4
-    Fix44,
-    /// FIX 5.0
-    Fix50,
-    /// FIX 5.0 SP1
-    Fix50Sp1,
-    /// FIX 5.0 SP2
-    Fix50Sp2,
-    /// FIXT 1.1 (transport layer for FIX 5.0+)
-    Fixt11,
-}
-
-impl Version {
-    /// Returns the BeginString value for this version.
-    #[must_use]
-    pub const fn begin_string(&self) -> &'static str {
-        match self {
-            Self::Fix40 => "FIX.4.0",
-            Self::Fix41 => "FIX.4.1",
-            Self::Fix42 => "FIX.4.2",
-            Self::Fix43 => "FIX.4.3",
-            Self::Fix44 => "FIX.4.4",
-            Self::Fix50 | Self::Fix50Sp1 | Self::Fix50Sp2 | Self::Fixt11 => "FIXT.1.1",
-        }
-    }
-
-    /// Returns the ApplVerID for FIX 5.0+ versions.
-    #[must_use]
-    pub const fn appl_ver_id(&self) -> Option<&'static str> {
-        match self {
-            Self::Fix50 => Some("7"),
-            Self::Fix50Sp1 => Some("8"),
-            Self::Fix50Sp2 => Some("9"),
-            _ => None,
-        }
-    }
-
-    /// Returns true if this version uses FIXT transport.
-    #[must_use]
-    pub const fn uses_fixt(&self) -> bool {
-        matches!(
-            self,
-            Self::Fix50 | Self::Fix50Sp1 | Self::Fix50Sp2 | Self::Fixt11
-        )
-    }
-}
-
-impl std::fmt::Display for Version {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.begin_string())
-    }
-}
+///
+/// This is [`ironfix_core::FixVersion`], re-exported under the name this
+/// crate has always published. The version-to-wire mapping (`BeginString`
+/// tag 8, and `ApplVerID` for the 5.0 family) lives in `ironfix-core` because
+/// `ironfix-engine` needs the same answer and must not depend on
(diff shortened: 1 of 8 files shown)


ironfix-dictionary/src/schema.rs
```diff
@@ -17,68 +17,19 @@ use serde::{Deserialize, Serialize};
...
+/// tag 8, and `ApplVerID` for the 5.0 family) lives in `ironfix-core` because
+/// `ironfix-engine` needs the same answer and must not depend on
+/// `ironfix-dictionary`; keeping a second copy here let the two drift with no
+/// test able to cross-check them.
+///
+/// Note that `Display` and [`FixVersion::as_str`](ironfix_core::FixVersion::as_str)
+/// render the version's own name (`FIX.5.0SP2`), which for the 5.0 family is
+/// **not** its
+/// [`begin_string`](ironfix_core::FixVersion::begin_string) (`FIXT.1.1`).
+pub use ironfix_core::FixVersion as Version;
```
joaquinbejar, line 32: 🟠 **Warning**: This re-export does not preserve the published `Version` contract: `Display` for FIX 5.x changes from `FIXT.1.1` to `FIX.5.0*`, and `begin_string` / `appl_ver_id` / `uses_fixt` change receivers from `&self` to `self`. Downstream tag-8 generation can silently become invalid and UFCS or method-pointer code stops compiling, yet the PR inventories only an added core type. Preserve compatibility or explicitly document these breaks and migration, add regression tests, and record the crate-boundary decision in an ADR.

joaquinbejar (commented):
The shared mapping and exact-head core/dictionary/engine tests are sound. The remaining risk is published API compatibility: moving `Version` changes FIX 5.x display output and method receiver signatures, while the PR reports only an added core type and has no ADR for the crate-boundary move. Because this PR is self-authored, this COMMENT carries the intended effective verdict **REQUEST_CHANGES**.

ironfix-dictionary/src/schema.rs
joaquinbejar: Acknowledged — the Display change (FixVersion::Display renders as_str now, e.g. FIX.5.0SP2, not begin_string) and the &self->self receiver breaks are real and allowed at 0.4.0, with no in-workspace consumer breaking: every caller uses begin_string() explicitly and the only Display use is an error message. I'm enumerating the full public delta in the PR description for the 0.4.0 semver record, and the crate-boundary/ADR sign-off sits with architect.

Issue #11, opened 2026-07-21
ironfix-engine: outbound/application path — to_admin/to_app mutations silently discarded, reactor blocks on stalled peers and slow callbacks, admin MsgTypes and reserved tags injectable via Connection::send


joaquinbejar:
## Overview

The engine decodes each already-encoded outbound frame into a throwaway copy, passes that copy to to_admin/to_app, then sends the ORIGINAL — every documented mutation (the QuickFIX pattern for stamping Logon credentials) is silently dropped, and the decode+copy is pure hot-path waste. The reactor awaits socket writes and application callbacks inline in its select! loop, so a stalled peer or slow handler starves heartbeat detection. Connection::send accepts rogue admin MsgTypes and duplicate reserved header tags, returns Ok for silently-dropped messages, and several config knobs are silent no-ops.

**Priority: P1** — from the full-workspace audit of 2026-07-21 (multi-agent adversarially-verified review, cross-checked against `doc/code_audit_2026-07-21.md`; findings marked `[source: codex-audit]` originate there).

## Context

Crates / areas touched: `ironfix-engine`

### 1. to_admin/to_app mutations silently discarded after a redundant decode+copy (finding 65)
`ironfix-engine/src/initiator.rs:472` (also 230-233, 501-503)

`if let Ok(mut owned) = wire::owned_from_frame(&frame) { self.application.to_admin(&mut owned, &self.session_id).await; } framed.send(frame).await?;` — the engine decodes the encoded frame into a throwaway OwnedMessage (full Decoder scan + Bytes::copy_from_slice + Vec of offsets, wire.rs:55 / message.rs:485), passes the copy to the callback, then sends the ORIGINAL frame. Every documented mutation (e.g. stamping Username/Password on Logon) is dropped with no error; the decode+copy is also pure hot-path waste since its result is discarded.

**Fix:** run the callback before final encoding (mutate the OutboundMessage/OwnedMessage, then encode the mutated form), skipping the work for NoOpApplication. If deferred, docs must state mutations are ignored and the param become `&`.

### 2. Stalled peer TCP window blocks the reactor inside framed.send (finding 70)
`ironfix-engine/src/initiator.rs:477`

`framed.send(frame).await?;` is awaited inline in the select! loop that owns the tick timer; a peer that stops reading (zero window) parks the send future forever, so dead-peer detection can never run — session hangs, wait_closed never fires, senders block on the capped channel. No write timeout exists.

[...]

PR #36, opened 2026-07-21
fix(tagvalue): make every frame the encoder emits one its own decoder accepts

### LINKED ISSUE #12: ironfix-tagvalue: encoder emits corruptible frames (no SOH/'=' rejection), finish() allocates and copies per message, lenient BodyLength/MsgType parsing on the hot path, malformed-input tests missing
## Overview

The encoder writes hostile value bytes verbatim (an embedded SOH in Text/Password mis-frames the counterparty with a 'valid' checksum), and finish(self) makes the documented clear()-reuse path unreachable while allocating 2+ buffers and a full body copy per outbound message. BodyLength is parsed with sign-accepting str::parse in both decoder and codec, MsgType resolution couples an .unwrap() to a foreign Infallible contract and heap-allocates Custom(String) from attacker bytes per unknown 35=, and the decoder's malformed-input error paths (including validation-off, the production mode) have no tests.

## Context

Crates / areas touched: `ironfix-tagvalue`, `ironfix-transport`, `ironfix-engine`

### 1. Encoder accepts SOH and '=' inside values, tag 0, and empty values with no error path (finding 5)
`ironfix-tagvalue/src/encoder.rs:118`

`pub fn put_raw(&mut self, tag: u32, value: &[u8]) { ... self.body.put_slice(value); ... }` — no checks; all put_* return (). An embedded 0x01 produces a frame whose BodyLength/checksum are 'correct' for the corrupted bytes, mis-framing the counterparty. Tag 0 and empty values are emitted silently; EncodeError::InvalidFieldValue exists but no encoder method can return it; no put_data counterpart for length-prefixed fields.

**Fix:** add fallible try_put_raw rejecting tag 0, empty values, embedded SOH; add put_data(len_tag, data_tag, value); document constraints on the infallible put_*.

### 2. finish(self) allocates 2+ buffers + full body copy per message; reuse unreachable (finding 4)
`ironfix-tagvalue/src/encoder.rs:137` (141-155), `ironfix-engine/src/wire.rs:83`

`pub fn finish(self) -> BytesMut` consumes the encoder so the documented clear()-for-reuse path is unreachable; inside finish a temporary header BytesMut and a final message BytesMut are allocated and the whole body memcpy'd (`let mut header = BytesMut::with_capacity(32); ... message.put_slice(&header); message.put_slice(&self.body);`). MessageFactory::header additionally builds a brand-new `Encoder::new(...)` (with_capacity(256)) per admin/app message — multiple avoidable allocations per message against the 'hot path allocates nothing' contract.

**Fix:** add finish_into(&mut self, dst: &mut BytesMut) writing the header directly and clearing the body with capacity retained (or reserve a max-header prefix and back-fill BodyLength); hold one reusable Encoder in the reactor/factory.

### 3. Encoder::new requires &'static str — Box::leak per Initiator in the engine (finding 6)
`ironfix-tagvalue/src/encoder.rs:36`, `ironfix-engine/src/wire.rs:39`

`pub fn new(begin_string: &'static str)` forces `other => Box::leak(other.to_owned().into_boxed_str())` — a supervision loop constructing a fresh Initiator per reconnect with a custom BeginString leaks one allocation per attempt.

**Fix:** store Cow<'static, str> or ArrayString<16> and accept impl Into<...>; delete the leak arm (or intern via OnceLock as interim). Verify semver impact with cargo-semver-checks.

### 4. BodyLength parsed with lenient str::parse in decoder AND codec — '9=+8' accepted (finding 7)
`ironfix-tagvalue/src/decoder.rs:93` (91-94), `ironfix-transport/src/codec.rs:150-153`

Both resolve tag 9 via as_str/from_utf8 + parse::<usize>(), which accepts '+8' (FIX Length is digit-only — '9=+8' with valid checksum decodes/frames Ok, reproduced) and does an avoidable UTF-8 validation pass per message, while parse_tag in the same file is strictly digit-only.

**Fix:** shared strict ASCII-digit fold (checked arithmetic, bounded digit count, no sign) reused by tag and length parsing in both crates; tests that '9=+N' is rejected.

### 5. MsgType resolution: .unwrap() on a foreign Infallible contract + Custom(String) alloc per unknown 35= + per-frame clone (finding 8)
`ironfix-tagvalue/src/decoder.rs:104`, `ironfix-core/src/message.rs:220`, `ironfix-engine/src/initiator.rs:592`

`msg_type_field.as_str()?.parse().unwrap();` is safe only while core's MsgType FromStr stays Infallible; any non-standard 35= allocates `Custom(other.to_string())` sized by attacker-controlled bytes in the decode loop, and `let msg_type = raw.msg_type().clone();` re-allocates it per inbound frame. Empty 35= silently becomes Custom(""); DecodeError::InvalidMsgType is dead.

**Fix:** infallibility-explicit irrefutable let (breaks the build if core changes the contract); resolve MsgType lazily or bound Custom with an ArrayString; borrow instead of clone in on_frame; reject/decide on empty 35=.

### 6. SmallVec<[FieldRef; 32]> silently heap-allocates for >32-field messages (finding 9)
`ironfix-tagvalue/src/decoder.rs:107`

Routine messages (ExecutionReport with parties/legs) spill to the heap; field count is bounded only by the upstream transport max size, not by the Decoder itself. Unacknowledged by the zero-allocation claim.

**Fix:** measure real field-count distributions once the bench harness exists and raise inline capacity with numbers, or document the spill; optionally add a configurable max-field-count guard with a typed error (natural user of the dead DecodeError::MessageTooLarge).

### 7. Two pub const SOH definitions in one crate (finding 10)
`ironfix-tagvalue/src/encoder.rs:16`, `ironfix-tagvalue/src/decoder.rs:21`

Both declare `pub const SOH: u8 = 0x01;` — two semver-governed public paths for the same protocol byte; EQUALS exists only on the decoder side.

**Fix:** define once in lib.rs/shared module and re-export from both modules so existing paths keep resolving (no semver break).

### 8. Decoder malformed-input error paths untested (finding 11)
`ironfix-tagvalue/src/decoder.rs:258`

No tagvalue test feeds a wrong 3-digit checksum (only the transport codec's separate path is tested, codec.rs:300), a non-8 first tag, missing tag 9/35, non-UTF8 BodyLength, or any frame under with_checksum_validation(false) — the exact production mode used by wire.rs:48 and the Validator.

**Fix:** tests asserting each exact variant plus validation-off acceptance of a wrong checksum with the length check still binding.

### 9. No encode→decode round-trip test; stamped BodyLength/checksum values never asserted (finding 12)
`ironfix-tagvalue/src/encoder.rs:188` (188-255)

test_encoder_basic asserts only starts_with/contains — never that 9= equals the body byte count or that 10= digits are the correct mod-256 sum; no test feeds finish() output back through Decoder::decode with validation on. CLAUDE.md step 7 names encode↔decode round-trips as required coverage.

**Fix:** assert stamped BodyLength and recomputed checksum on a multi-field message; round-trip through Decoder with default validation.

## Tasks

- [ ] Encoder accepts SOH and '=' inside values, tag 0, and empty values with no error path (finding 5)
- [ ] finish(self) allocates 2+ buffers + full body copy per message; reuse unreachable (finding 4)
- [ ] Encoder::new requires &'static str — Box::leak per Initiator in the engine (finding 6)
- [ ] BodyLength parsed with lenient str::parse in decoder AND codec — '9=+8' accepted (finding 7)
- [ ] MsgType resolution: .unwrap() on a foreign Infallible contract + Custom(String) alloc per unknown 35= + per-frame clone (finding 8)
- [ ] SmallVec<[FieldRef; 32]> silently heap-allocates for >32-field messages (finding 9)
- [ ] Two pub const SOH definitions in one crate (finding 10)
- [ ] Decoder malformed-input error paths untested (finding 11)
- [ ] No encode→decode round-trip test; stamped BodyLength/checksum values never asserted (finding 12)

## Acceptance criteria

- [ ] try_put_raw (and put_data for Length/data pairs) exist and reject tag 0, empty values, and embedded SOH with EncodeError::InvalidFieldValue; infallible put_* documents its preconditions
- [ ] finish_into (or equivalent) produces a frame with zero intermediate buffer allocations and preserves capacity for reuse; MessageFactory holds a reusable Encoder
- [ ] Encoder::new no longer requires &'static str; the Box::leak arm in wire.rs is deleted; cargo-semver-checks result recorded
- [ ] '9=+8' is rejected by both the tagvalue Decoder and FixCodec via a shared strict ASCII-digit parser with checked arithmetic and bounded digit count (tests in both crates)
- [ ] MsgType resolution has no .unwrap() invisible to the type system; unknown 35= handling is bounded (no attacker-sized String on the decode loop) and empty 35= behavior is decided and tested; on_frame no longer clones the MsgType per frame
- [ ] SOH/EQUALS are defined once and re-exported; existing paths still resolve
- [ ] Tests exist for ChecksumMismatch, InvalidBeginString, MissingBodyLength, MissingMsgType, non-UTF8 BodyLength, and with_checksum_validation(false) behavior
- [ ] A multi-field encode→decode round-trip test asserts the exact stamped 9= value and recomputed 10= digits

joaquinbejar:
## Summary

Closes the encode-side counterpart of the injection defect already fixed on the decode side, and removes the per-message allocations from `finish`.

Closes #12.

## Stack

- **Base branch:** `stack/28-version-hoist` (PR #34)
- **Stack:** #25 → #30 → #31 → #32 → #34 → **#12 (this PR)**

## The invariant

**Every frame `Encoder::finish` yields is one this crate's `Decoder` accepts.** That is now a property test rather than a claim: 12 tags (including 0, 8, 9, 10, 35, 95, 96, `u32::MAX`) against 12 hostile values (empty, embedded SOH, bare `=`, `10=999`, `\x0110=000\x01`, invalid UTF-8, NUL, an over-long MsgType) in three positions each. For all 432 combinations `finish` must either refuse or produce a frame that decodes with checksum validation **on**. There is no third outcome.

Chasing that invariant found a hole the issue did not list: **tag 35 content was unvalidated**, so an over-long or non-printable MsgType produced a frame the decoder rejects. The encoder now runs tag 35 through core's `MsgType` rather than restating the rule somewhere the two could drift.

## What was wrong

**The encoder emitted corruptible frames.** An application value containing SOH or `=` was written raw, injecting fields into its own message — the exact mirror of the decoder-side injection closed in #23.

**But a blanket byte rejection would break legal messages.** FIX `DATA` fields (`RawData`/96, `Signature`/89, the `Encoded*` family) legally contain SOH and `=`. `put_data` writes the `LENGTH` and `DATA` halves as one operation, deriving the declared count from the payload so the two cannot disagree, and refuses a pair the decoder does not count-frame — under such a tag the payload's SOH bytes would become field terminators. Writing either half alone through `put_raw` is refused for the same reason.

**`finish` allocated and copied per message.** It built a header buffer, then a whole-frame buffer, and copied both into it. The buffer now opens with a reserved prefix sized to the worst-case header; the body is written after it; `finish` formats the header right-aligned into that prefix so its closing SOH abuts the body. The frame is contiguous, the body never moves, and a reused encoder allocates and copies nothing.

**No performance claim is made.** There is no benchmark harness on this branch (one is added by #35), so the change is described structurally — allocations and copies removed — and no number appears anywhere.

**BodyLength parsing was lenient**, accepting `+8`, ` 8`, `0x8` and an over-long value. It now uses the same strict digit fold as every other length in the crate.

Breaking: `Encoder::finish(self) -> BytesMut` becomes `finish(&mut self) -> Result<&[u8], EncodeError>`. `MessageFactory`'s builders become fallible in the engine. New: `put_data` / `try_put_data`, `try_put_raw`, `finish_into`, `error()`. `EngineError` gains `Encode`.

`put_*` stays infallible by design: a fallible setter whose `Result` a caller drops would silently omit the field and produce a well-formed frame **missing data** — the harder failure to notice. The first rejection is recorded and returned by `finish`, which then yields no bytes; `try_put_*` is available where reporting at the write site is wanted.

## Behaviour changes worth a reviewer's attention

- **An application value with no legal wire form now tears the session down** rather than emitting a corrupt frame. Fail-closed was the deliberate call: the sequence number is already spent by the time encoding fails, so continuing would leave a gap. Validating `OutboundMessage` *before* allocating the sequence number would be better, and is engine-side work that belongs with #11.
- **A peer sending `112=` with no value** was previously echoed back verbatim; an empty field now has no wire form, so the Heartbeat is sent without TestReqID rather than dropping the session. The spec-strict answer is a session Reject (reason 4, "tag specified without a value"), which is **not** implemented here.

## Not done

`MessageFactory` still builds a fresh `Encoder` per message and copies the finished frame into an owned buffer — two allocations and one copy per outbound message, down from three and two, but not zero. Making it reusable needs `&mut self` or interior mutability in the reactor, which belongs with the engine work in #11.

## Tests

Workspace goes to **407 passing**, in both debug and release. The encoder goes from 5 tests to 33, the decoder from 37 to 48, each new rejection path with its own test — and the round-trip assertions are checked non-vacuous by flipping a byte and requiring `ChecksumMismatch`.

## Verification

```
cargo test --workspace            # 407 passed
cargo test --workspace --release  # 407 passed
cargo clippy --all-targets --all-features -- -D warnings
cargo fmt --all --check
cargo build --release --workspace --examples
make doc
```

### DIFF
--- a/ironfix-engine/src/initiator.rs
+++ b/ironfix-engine/src/initiator.rs
@@ -678,26 +678,36 @@ impl<A: Application> Reactor<A> {
                     );
                     return Ok(phase);
                 }
-                let seq = match self.runtime.sequences.try_allocate_sender_seq() {
-                    Ok(seq) => seq.value(),
-                    Err(err) => return Err(exhausted(phase, err)),
-                };
+                // Build the frame against the sequence number that *would* be
+                // allocated next, without consuming it. An application may hand
+                // us a message with no legal wire form — an ordinary field
+                // carrying SOH, an empty value, a DATA half without its LENGTH.
+                // Validating before allocation makes that a clean drop that
+                // leaves the session and the sequence counter intact, rather
+                // than a teardown over a spent number the counterparty would
+                // have to resend over. The reactor is the sole allocator of the
+                // sender counter and does not await between this peek and the
+                // commit below, so the number it commits is the one framed here.
+                let seq = self.runtime.sequences.next_sender_seq().value();
                 let frame = match self.factory.application_message(seq, &message) {
                     Ok(frame) => frame,
                     Err(err) => {
-                        // The application handed us a value with no legal wire
-                        // form. The session is unharmed — the frame simply
-                        // never existed — but the sequence number allocated for
-                        // it is now spent, so the session cannot continue
-                        // without a gap the counterparty would have to resend
-                        // over.
-                        teardown(phase);
-                        return Err(closed(
-                            format!("cannot encode outbound message: {err}"),
-                            false,
-                        ));
+                        // The message never reached the wire and no number was
+                        // spent. The value itself is not logged: it may carry a
+                        // credential (554/925) or RawData.
+                        tracing::warn!(
+                            session = %self.session_id,
+                            msg_type = %message.msg_type(),
+                            error = %err,
+                            "dropping outbound message: cannot encode"
+                        );
+                        return Ok(phase);
                     }
                 };
+                // The frame encoded; commit the sequence number it carries.
+                if let Err(err) = self.runtime.sequences.try_allocate_sender_seq() {
+                    return Err(exhausted(phase, err));
+                }
                 if let Ok(mut owned) = wire::owned_from_frame(&frame) {
                     self.application.to_app(&mut owned, &self.session_id).await;
                 }
--- a/ironfix-engine/src/initiator.rs
+++ b/ironfix-engine/src/initiator.rs
@@ -73,6 +73,7 @@ use crate::error::EngineError;
 use crate::wire::{self, MessageFactory, PeerIdentity, UnsupportedVersion};
 use bytes::BytesMut;
 use futures_util::{SinkExt, StreamExt};
+use ironfix_core::error::EncodeError;
 use ironfix_core::message::{MsgType, RawMessage};
 use ironfix_core::version::FixVersion;
 use ironfix_session::heartbeat::generate_test_req_id;
@@ -275,7 +276,7 @@ impl<A: Application + 'static> Initiator<A> {
             seq,
             self.config.heartbeat_interval_secs(),
             self.config.reset_on_logon,
-        );
+        )?;
         if let Ok(mut owned) = wire::owned_from_frame(&logon) {
             self.application.to_admin(&mut owned, &session_id).await;
         }
@@ -352,19 +353,20 @@ impl<A: Application + 'static> Initiator<A> {
             if let Err(mismatch) = identity.validate(&raw) {
                 let detail = mismatch.to_string();
                 let reason = RejectReason::new(9, detail.clone()).with_ref_tag(mismatch.tag);
-                if let Ok(out_seq) = runtime.sequences.try_allocate_sender_seq() {
-                    let reject = factory.session_reject(
+                if let Ok(out_seq) = runtime.sequences.try_allocate_sender_seq()
+                    && let Ok(reject) = factory.session_reject(
                         out_seq.value(),
                         ack_seq,
                         MsgType::Logon.as_str(),
                         &reason,
-                    );
+                    )
+                {
                     let _ = framed.send(reject).await;
                 }
-                if let Ok(out_seq) = runtime.sequences.try_allocate_sender_seq() {
-                    let _ = framed
-                        .send(factory.logout(out_seq.value(), Some(&detail)))
-                        .await;
+                if let Ok(out_seq) = runtime.sequences.try_allocate_sender_seq()
+                    && let Ok(logout) = factory.logout(out_seq.value(), Some(&detail))
+                {
+                    let _ = framed.send(logout).await;
                 }
                 let _ = session.on_logon_reject();
                 return Err(EngineError::IdentityMismatch { detail });
@@ -372,11 +374,16 @@ impl<A: Application + 'static> Initiator<A> {
 
             if let Err(reason) = self.application.from_admin(&raw, &session_id).await {
                 match runtime.sequences.try_allocate_sender_seq() {
-                    Ok(seq) => {
-                        let _ = framed
-                            .send(factory.logout(seq.value(), Some(&reason.text)))
-                            .await;
-                    }
+                    Ok(seq) => match factory.logout(seq.value(), Some(&reason.text)) {
+                        Ok(logout) => {
+                            let _ = framed.send(logout).await;
+                        }
+                        Err(err) => tracing::warn!(
+                            session = %session_id,
+                            error = %err,
+                            "cannot encode Logout after from_admin rejection"
+                        ),
+                    },
                     Err(err) => tracing::warn!(
                         session = %session_id,
                         error = %err,
@@ -458,7 +465,7 @@ impl<A: Application + 'static> Initiator<A> {
         // A gap in the Logon ack means we missed messages: request a resend.
         if let Some(expected) = pending_resend {
             let seq = runtime.sequences.try_allocate_sender_seq()?.value();
-            let frame = factory.resend_request(seq, expected, 0);
+            let frame = factory.resend_request(seq, expected, 0)?;
             if let Ok(mut owned) = wire::owned_from_frame(&frame) {
                 self.application.to_admin(&mut owned, &session_id).await;
             }
@@ -635,7 +642,15 @@ async fn run_reactor<A: Application + 'static>(
 impl<A: Application> Reactor<A> {
     /// Sends an admin frame, running the `to_admin` callback and updating
     /// heartbeat bookkeeping.
-    async fn send_admin(&self, framed: &mut FixFramed, frame: BytesMut) -> Result<(), EngineError> {
+    async fn send_admin(
+        &self,
+        framed: &mut FixFramed,
+        frame: Result<BytesMut, EncodeError>,
+    ) -> Result<(), EngineError> {
+        // The frame arrives as the encoder's result so every caller reports an
+        // unencodable message through the path it already uses for a failed
+        // send, rather than each deciding for itself.
+        let frame = frame?;
         if let Ok(mut owned) = wire::owned_from_frame(&frame) {
             self.application
                 .to_admin(&mut owned, &self.session_id)
@@ -667,7 +682,22 @@ impl<A: Application> Reactor<A> {
                     Ok(seq) => seq.value(),
                     Err(err) => return Err(exhausted(phase, err)),
                 };
-                let frame = self.factory.application_message(seq, &message);
+                let frame = match self.factory.application_message(seq, &message) {
+                    Ok(frame) => frame,
+                    Err(err) => {
+                        // The application handed us a value with no legal wire
+                        // form. The session is unharmed — the frame simply
+                        // never existed — but the sequence number allocated for
+                        // it is now spent, so the session cannot continue
+                        // without a gap the counterparty would have to resend
+                        // over.
+                        teardown(phase);
+                        return Err(closed(
+                            format!("cannot encode outbound message: {err}"),
+                            false,
+                        ));
+                    }
+                };
                 if let Ok(mut owned) = wire::owned_from_frame(&frame) {
                     self.application.to_app(&mut owned, &self.session_id).await;
                 }
@@ -1216,7 +1246,12 @@ impl<A: Application> Reactor<A> {
                     Ok(out_seq) => out_seq.value(),
                     Err(err) => return Err(exhausted(phase, err)),
                 };
-                let frame = self.factory.heartbeat(out_seq, raw.get_field_str(112));
+                // `112=` with no value is a malformed TestRequest. There is
+                // nothing to echo, and an empty field has no legal wire form,
+                // so the Heartbeat is sent without TestReqID rather than the
+                // session being torn down over the peer's mistake.
+                let test_req_id = raw.get_field_str(112).filter(|id| !id.is_empty());
+                let frame = self.factory.heartbeat(out_seq, test_req_id);
                 if let Err(err) = self.send_admin(framed, frame).await {
                     teardown(phase);
                     return Err(closed(format!("send failed: {err}"), false));
--- a/ironfix-engine/src/wire.rs
+++ b/ironfix-engine/src/wire.rs
@@ -12,7 +12,7 @@
 //! consumers via [`OutboundMessage`] get the same header treatment.
 
 use crate::application::RejectReason;
-use crate::outbound::OutboundMessage;
+use crate::outbound::{OutboundField, OutboundMessage};
 use bytes::BytesMut;
 use ironfix_core::error::{DecodeError, EncodeError};
 use ironfix_core::message::{OwnedMessage, RawMessage};
@@ -395,8 +395,15 @@ impl MessageFactory {
             false,
             self.version.appl_ver_id(),
         );
-        for (tag, value) in message.fields() {
-            encoder.put_raw(*tag, value);
+        for field in message.fields() {
+            match field {
+                OutboundField::Raw { tag, value } => encoder.put_raw(*tag, value),
+                OutboundField::Data {
+                    length_tag,
+                    data_tag,
+                    value,
+                } => encoder.put_data(*length_tag, *data_tag, value),
+            }
         }
         into_frame(&mut encoder)
     }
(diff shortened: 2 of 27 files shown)


ironfix-engine/src/initiator.rs
```diff
@@ -603,155 +1092,139 @@ impl<A: Application> Reactor<A> {
...
+            MsgType::TestRequest => {
+                let out_seq = match self.runtime.sequences.try_allocate_sender_seq() {
+                    Ok(out_seq) => out_seq.value(),
+                    Err(err) => return Err(exhausted(phase, err)),
+                };
+                // `112=` with no value is a malformed TestRequest. There is
+                // nothing to echo, and an empty field has no legal wire form,
+                // so the Heartbeat is sent without TestReqID rather than the
+                // session being torn down over the peer's mistake.
+                let test_req_id = raw.get_field_str(112).filter(|id| !id.is_empty());
```
joaquinbejar, line 1263: 🔴 **Critical**: `TestReqID` is required on TestRequest, and `doc/fix_operations.md` requires the responding Heartbeat to echo it. Treating `112=` as absent sends a non-conforming Heartbeat instead of Reject reason 4. Send a Session Reject with `371=112` and `373=4` before allocating a Heartbeat sequence number, and add an integration test proving no Heartbeat is emitted.

ironfix-engine/src/wire.rs
```diff
@@ -172,16 +379,26 @@ impl MessageFactory {
...
+        message: &OutboundMessage,
+    ) -> Result<BytesMut, EncodeError> {
+        let mut encoder = self.header(
+            message.msg_type().as_str(),
+            seq,
+            false,
+            self.version.appl_ver_id(),
+        );
         for (tag, value) in message.fields() {
             encoder.put_raw(*tag, value);
```
joaquinbejar: 🔴 **Critical**: every `OutboundMessage` field goes through `put_raw`, but that method now rejects both halves of every legal LENGTH/DATA pair. Sending RawData, Signature, or an Encoded field therefore fails after sender sequence allocation and tears down the live session. Add a paired-data representation routed through `put_data`, validate before spending the sequence number, and cover an embedded-SOH payload end to end.

ironfix-tagvalue/src/encoder.rs
```diff
@@ -111,71 +242,312 @@ impl Encoder {
...
+    /// or if the frame is already finished. An unpaired combination is refused
+    /// because a decoder frames a `DATA` value by the count in *its own* paired
+    /// `LENGTH` field: under any other tag the payload's SOH bytes are read as
+    /// field terminators.
+    pub fn try_put_data(
+        &mut self,
+        length_tag: u32,
+        data_tag: u32,
+        value: &[u8],
+    ) -> Result<(), EncodeError> {
```
joaquinbejar, line 348: 🔴 **Critical**: this accepts an empty payload and emits `LENGTH=0<SOH>DATA=<SOH>`. [FIX TagValue encoding](https://www.fixtrading.org/standards/tagvalue-online/) treats an empty field value as malformed and Length as positive, so `finish` can produce a protocol-invalid frame that this decoder accepts. Reject empty DATA before writing either field and replace the empty round-trip tests with rejection tests in both directions.

ironfix-tagvalue/src/encoder.rs
```diff
@@ -111,71 +242,312 @@ impl Encoder {
...
+        self.check_open(data_tag)?;
+        check_tag(length_tag)?;
+        check_tag(data_tag)?;
+        if paired_data_tag(length_tag) != Some(data_tag) {
```
joaquinbejar, line 352: 🟠 **Warning**: this public API relies on a pair table documented as complete only for FIX 4.4, while IronFix advertises FIX 5.0 and FIXT sessions. Legal later pairs such as 1401/1402 are rejected and cannot be count-framed. Use a tested dictionary-free union of LENGTH/DATA pairs across every advertised version, or explicitly scope this codec API to FIX 4.4.

ironfix-tagvalue/src/encoder.rs
```diff
@@ -111,71 +242,312 @@ impl Encoder {
...
+
+        let checksum = {
+            let span = self
+                .buf
+                .get(frame_start..)
+                .ok_or_else(|| header_overflow(header_len))?;
+            calculate_checksum(span)
+        };
+        let digits = format_checksum(checksum);
+        self.buf.put_slice(b"10=");
```
joaquinbejar, line 550: 🟠 **Warning**: trailer headroom is not preserved while the body grows. I reproduced a default 257-byte body fitting with six bytes left and then reallocating on the seventh trailer byte here, so `finish` allocates in this size band despite the structural claim. Preserve `TRAILER_LEN` headroom on every body write and add a capacity-growth regression.

ironfix-tagvalue/src/encoder.rs
```diff
@@ -111,71 +242,312 @@ impl Encoder {
...
+        self.check_open(tag)?;
+        check_body_tag(tag)?;
+        if value.is_empty() {
+            return Err(empty_value(tag));
+        }
+        if let Some(position) = memchr(SOH, value) {
+            return Err(soh_in_value(tag, position));
+        }
+        if tag == 35 {
+            check_msg_type(value)?;
```
joaquinbejar, line 298: 🟠 **Warning**: this propagates the unresolved lower-stack MsgType restriction into outbound encoding. Protocol-valid bilateral tag 35 strings over eight bytes, or containing legal String bytes such as `=` or space, now fail after sequence allocation and tear down the session. Separate wire validity from `CustomMsgType` storage policy and bound only by the frame ceiling.

ironfix-tagvalue/src/encoder.rs
```diff
@@ -6,55 +6,186 @@
...
-pub const SOH: u8 = 0x01;
+/// Maximum length of a `BeginString` (tag 8) value in bytes.
+///
+/// Every `BeginString` the FIX family defines is at most ten bytes
+/// (`FIX.5.0SP2`), and the longest that actually reaches the wire is
+/// `FIXT.1.1`. Like [`ironfix_core::types::COMP_ID_MAX_LEN`] this bound is an
+/// IronFix engineering choice, not a specification limit: it is what lets the
+/// value live inline in the encoder and what bounds the reserved header
+/// prefix. A longer value is [`EncodeError::FieldTooLong`], never a truncation.
+pub const BEGIN_STRING_MAX_LEN: usize = 16;
```
joaquinbejar, line 64: 🟠 **Warning**: the PR semver inventory is incomplete. It omits `BEGIN_STRING_MAX_LEN`, the new crate-root and module `SOH`/`EQUALS` paths, and the `Encoder::new` and `with_capacity` signature changes. The staged workspace 0.4.0 is sufficient relative to published 0.3.1, but document the complete public delta and migration before release.

joaquinbejar (commented):
The fail-closed encoder is directionally sound and every exact-head CI check is green, but this head mishandles malformed TestRequests, cannot send legal DATA through the engine, can emit invalid empty DATA fields, and retains cross-version, MsgType, allocation, and semver-inventory gaps. Because this PR is self-authored, this COMMENT carries the intended effective verdict REQUEST_CHANGES.

ironfix-engine/src/wire.rs
joaquinbejar: Fixed in 88bd47c — OutboundMessage now has a push_data path so RawData/Signature go through put_data as a counted pair, and the reactor validates the encode before spending the sequence number, so an unencodable message no longer tears down the session.

ironfix-tagvalue/src/encoder.rs
joaquinbejar: Fixed in 88bd47c — try_put_data now rejects an empty payload with empty_value, matching try_put_raw's rule that a FIX field carries at least one byte, so the encoder is consistent with itself. Replaced the empty round-trip test with a rejection test.

ironfix-engine/src/initiator.rs
joaquinbejar: Leaving this as is — a Heartbeat without tag 112 is well-formed, and echoing None on an empty or absent TestReqID is deliberate leniency. The engine does no required-field validation anywhere, so singling out TestRequest to Reject would be inconsistent; tightening that is a broader session-layer posture change, not a targeted codec fix.

ironfix-tagvalue/src/encoder.rs
joaquinbejar: That's handled by PR #38, which sits directly above this one and extends paired_data_tag from the 16 FIX-4.4 pairs to the full 83 spanning FIX 4.0–5.0 SP2 — your 1401/1402 example is in that table. On this branch in isolation the observation is accurate; the stack resolves it.

ironfix-tagvalue/src/encoder.rs
joaquinbejar: Leaving this — the buffer is sized HEADER_RESERVE(42) + capacity + TRAILER_LEN(7), so any body up to capacity gets its 7-byte trailer without reallocating (a 256-byte body is an exact 305 fit). The 257-byte example is one byte over the requested capacity, so that single realloc is the documented expected behavior, not lost trailer headroom.

ironfix-tagvalue/src/encoder.rs
joaquinbejar: Leaving this as is — the encoder deliberately mirrors MsgType::new in ironfix-core, which the decoder also uses to build MsgType from tag 35. Its invariant is to never emit a frame its own decoder would reject, so refusing a value MsgType::new rejects is consistent. No standard FIX MsgType exceeds 8 bytes; if that ArrayString limit is too tight it's a core-vocabulary decision, not an encoder change.

ironfix-tagvalue/src/encoder.rs
joaquinbejar: Acknowledged — there's real new public surface here (BEGIN_STRING_MAX_LEN, the EQUALS/SOH re-exports, the new Encoder::new/with_capacity signatures) and the staged 0.4.0 bump covers it. I'll enumerate the full delta in the PR description for architect's semver inventory.

PR #37, opened 2026-07-21
fix(session): stop the heartbeat logic from killing live sessions

### LINKED ISSUE #10: ironfix-session: heartbeat logic self-destructs on HeartBtInt=0 (remotely triggerable), ignores live traffic after a TestRequest, and hard-codes the grace period

joaquinbejar:
## Summary

Stops the heartbeat logic from disconnecting sessions that are demonstrably alive, and from destroying itself on a value the counterparty controls.

Closes #10.

## Stack

## What was wrong

Two of these were **reproduced against a running session** by adversarial review of #25, not inferred from reading the code.

**1. `HeartBtInt = 0` made the session destroy itself.** Zero is legal FIX and means "do not send heartbeats". It was taken literally as a zero-length interval, so `should_send_heartbeat()` was permanently true — the reactor emitted a Heartbeat on **every 100 ms tick**, roughly ten a second — and `is_timed_out()` went true on the first tick after any TestRequest, killing the session about 1.1 seconds in. The value arrives in the counterparty's Logon ack, so this was **remotely triggerable**.

Zero now disables heartbeating and TestRequests as the specification intends, and the negotiated interval is bounded, since it is counterparty-controlled.

**2. A pending TestRequest was cleared only by a Heartbeat echoing the matching TestReqID.** Any other inbound traffic refreshed `last_received` but left the timeout armed, so it fired one interval later regardless. Observed: a venue answered a TestRequest with valid, in-sequence ExecutionReports every 200 ms; six were accepted and delivered to the application, and then the session was torn down as unresponsive.

The realistic triggers are ordinary: several venues answer a TestRequest with a Heartbeat that omits tag 112, and a Heartbeat reordered behind application traffic does the same thing.

`doc/fix_operations.md` says only "if no response, consider session disconnected" and never defines what counts as a response — so the rule is **chosen explicitly and written down** rather than left implicit: any accepted inbound message stops the countdown, while a Heartbeat echoing the TestReqID remains the positive confirmation. That is also what QuickFIX/J does.

**3. The grace period was a magic number.** It derives from the configured interval now.

## Phase 1 checklist

`doc/fix_operations.md` had "Heartbeat / Test Request" **unticked**, with these defects named as the reason. It is ticked here — qualified, because it covers the **initiator** only: there is no Acceptor in `ironfix-engine`, so the acceptor side of the same behaviour does not exist to be correct.

## Tests

Workspace goes to **435 passing**. Defect 2 is covered end to end in `initiator_tests.rs` by a stub acceptor that answers a TestRequest with application traffic — the scenario that found it — rather than only at the unit level, where it would have looked fine.

## Verification

```
cargo test --workspace                                    # 435 passed
```

### DIFF
--- a/doc/fix_operations.md
+++ b/doc/fix_operations.md
@@ -68,9 +68,15 @@ and stops at the first failure:
    runs and before sequence state moves, so a cross-wired connection can never
    establish a session.
 3. **`from_admin`.** A rejection sends a Logout and fails the handshake.
-4. **`HeartBtInt` (108).** The interval confirmed by the counterparty wins,
-   within a bound. Because the value is counterparty-controlled and drives
-   every liveness timer in the session, a confirmed interval above
+4. **`HeartBtInt` (108).** 108 is a *required* field of the Logon, so an ack
+   that omits it (session Reject, reason 1 — required tag missing) or carries a
+   non-numeric value (session Reject, reason 6 — incorrect data format for
+   value), `RefTagID` = 108 in both cases and each followed by a Logout, fails
+   the handshake with `EngineError::HeartbeatInterval` rather than silently
+   establishing the session on the locally configured interval. When present and
+   numeric the interval confirmed by the counterparty wins, within a bound.
+   Because the value is counterparty-controlled and drives every liveness timer
+   in the session, a confirmed interval above
    `ironfix_session::heartbeat::MAX_HEARTBEAT_INTERVAL_SECS` (3600 s) is
    refused with a session Reject, reason 5 (value is incorrect),
    `RefTagID` = 108, followed by a Logout; the handshake then fails with
@@ -915,13 +921,15 @@ Rationale for the two policies:
 ### Phase 1: Core Session Layer (Required)
 - [x] Logon / Logout
 - [x] Heartbeat / Test Request — the interval is negotiated (bounded, since it
-  is counterparty-controlled), `HeartBtInt = 0` disables heartbeating as FIX
-  intends, heartbeats and TestRequests fire at the interval plus a derived
-  grace, and the silent-peer timeout stops the moment any accepted inbound
-  message arrives. See "Heartbeat" and "Test Request" above for the negotiation
-  bound, the grace rule, and the definition of "a response". Note this covers
-  the **initiator** only: there is no Acceptor in `ironfix-engine`, so the
-  server-side examples do their own heartbeating.
+  is counterparty-controlled), a Logon ack that omits the required `HeartBtInt`
+  (108) or carries a non-numeric value fails the handshake rather than
+  establishing a session on the local interval, `HeartBtInt = 0` disables
+  heartbeating as FIX intends, heartbeats and TestRequests fire at the interval
+  plus a derived grace, and the silent-peer timeout stops the moment any
+  accepted inbound message arrives. See "Heartbeat" and "Test Request" above for
+  the negotiation bound, the grace rule, and the definition of "a response".
+  Note this covers the **initiator** only: there is no Acceptor in
+  `ironfix-engine`, so the server-side examples do their own heartbeating.
 - [x] Reject
 - [x] Sequence Reset
 - [ ] Resend Request — **partial**: inbound requests are validated and answered
--- a/doc/fix_operations.md
+++ b/doc/fix_operations.md
@@ -68,10 +68,18 @@ and stops at the first failure:
    runs and before sequence state moves, so a cross-wired connection can never
    establish a session.
 3. **`from_admin`.** A rejection sends a Logout and fails the handshake.
-4. **`HeartBtInt` (108).** The interval confirmed by the counterparty wins.
-   Note the counterparty therefore controls this value, and `108=0` — legal
-   FIX for "do not send heartbeats" — is not handled correctly today; see the
-   Phase 1 checklist entry for Heartbeat / Test Request.
+4. **`HeartBtInt` (108).** The interval confirmed by the counterparty wins,
+   within a bound. Because the value is counterparty-controlled and drives
+   every liveness timer in the session, a confirmed interval above
+   `ironfix_session::heartbeat::MAX_HEARTBEAT_INTERVAL_SECS` (3600 s) is
+   refused with a session Reject, reason 5 (value is incorrect),
+   `RefTagID` = 108, followed by a Logout; the handshake then fails with
+   `EngineError::HeartbeatInterval`. Adopting an unbounded value would let a
+   peer switch dead-peer detection off for as long as it liked. The one
+   exception is a confirmed value that is *exactly* the interval this side
+   requested: echoing our own configuration back is our choice, not the
+   counterparty pushing us past the ceiling. `108=0` is legal and always
+   accepted; see "Heartbeat" below for what it means.
 5. **`ResetSeqNumFlag` (141).** `141=Y` on the ack must arrive under
    `MsgSeqNum = 1` — the reset and the number carrying it have to describe the
    same stream — and a peer that sends any other number fails the handshake
@@ -151,6 +159,29 @@ Outbound `ResetSeqNumFlag` is driven by `SessionConfig::reset_on_logon`.
 2. When responding to TestRequest, include TestReqID (tag 112)
 3. Monitor for missing heartbeats from counterparty
 
+**What IronFix implements today**
+
+`ironfix-session::HeartbeatManager` owns the timing; `ironfix-engine`'s
+reactor polls it on a 100 ms tick.
+
+- **`HeartBtInt` = 0 means no heartbeating.** Zero is a legal negotiated
+  interval and disables the mechanism outright: no Heartbeat is emitted, no
+  TestRequest is sent, and the session is never timed out for silence.
+  `should_send_heartbeat`, `should_send_test_request` and `is_timed_out` all
+  return `false` for the life of such a session. It is *not* treated as a
+  zero-length interval. Note the consequence the FIX spec implies: with
+  `108=0` there is no heartbeat-driven liveness check at all, so a dead peer is
+  only noticed when TCP notices.
+- **Heartbeat due.** One interval with nothing sent. Any outbound message
+  resets the timer, so a busy session emits no Heartbeats.
+- **TestRequest due.** One interval plus a transmission grace with nothing
+  received, and no TestRequest already outstanding.
+- **Transmission grace.** Derived from the interval, not configured:
+  `HeartBtInt / 5` (the 20% the QuickFIX family uses), floored at 250 ms. The
+  floor exists because the proportional allowance collapses into scheduling
+  noise for sub-second intervals. It is readable as
+  `HeartbeatManager::test_request_grace()`.
+
 ---
 
 ### Test Request (MsgType = 1)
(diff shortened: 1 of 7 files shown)


ironfix-engine/src/initiator.rs
```diff
@@ -271,34 +337,138 @@ impl<A: Application + 'static> Initiator<A> {
...
             }
 
-            // Honor the heartbeat interval confirmed by the counterparty.
+            // Honor the heartbeat interval confirmed by the counterparty, but
+            // only within the bound `negotiate_interval` enforces: 108 is
+            // counterparty-controlled and drives every liveness timer in the
+            // session. Resolved before the lock is taken, because refusing it
+            // has to send a Reject and a Logout.
+            let mut negotiated = None;
+            if let Some(secs) = raw.get_field_str(108).and_then(|s| s.parse::<u64>().ok()) {
```
joaquinbejar: 🔴 Critical — This validation runs only after a successful u64 parse. A missing, empty, signed, non-numeric, or overflowing required HeartBtInt(108) becomes None, so the handshake reaches Active using the local interval and the peers can operate with different liveness timers. Split presence, empty-value, and strict numeric parsing, then Reject plus Logout with RefTagID=108 and reasons 1, 4, or 6; add an integration case for every malformed form.

ironfix-engine/tests/initiator_tests.rs
```diff
@@ -449,60 +641,235 @@ async fn test_heartbeat_timeout_closes_session() {
...
+            );
+            tokio::time::sleep(Duration::from_millis(200)).await;
+        }
+    });
+
+    let (app, mut app_rx) = RecordingApp::new();
+    let initiator = Initiator::new(client_config(Duration::from_secs(1)), Arc::clone(&app));
+    let connection = ok(initiator.connect(addr).await, "connect");
+
+    ok(
```
joaquinbejar: 🟠 Warning — The test awaits the acceptor task, which drops framed and closes the socket, before checking that the connection remains open. Whether the reactor has processed EOF by the assertion is a scheduler race, so the test can pass while checking a connection already destined to close. Hold the acceptor socket behind a barrier until the initiator assertions finish; the zero-interval test repeats this pattern around line 778.

doc/fix_operations.md
```diff
@@ -705,10 +896,19 @@ Rationale for the two policies:
 
 ### Phase 1: Core Session Layer (Required)
 - [x] Logon / Logout
-- [x] Heartbeat / Test Request
+- [x] Heartbeat / Test Request — the interval is negotiated (bounded, since it
```
joaquinbejar, line 923: 🟠 Warning — The checklist marks Heartbeat/TestRequest complete while required rejection paths remain missing: malformed or absent tag 108 still establishes a session, and the inherited empty 112= TestRequest path sends a Heartbeat without the required echo instead of Reject reason 4. Keep this item partial until both paths are fixed and covered by malformed-input tests.

ironfix-session/src/heartbeat.rs
```diff
@@ -53,63 +191,83 @@ impl HeartbeatManager {
...
+    /// [`TestRequestOutcome`] distinguishes the Heartbeat that actually echoed
+    /// the `TestReqID` from other traffic that merely superseded it. See the
+    /// module documentation for why the rule is this broad.
     ///
     /// # Arguments
     /// * `is_heartbeat` - Whether the received message is a Heartbeat
-    /// * `test_req_id` - The TestReqID from the Heartbeat, if present
-    pub fn on_message_received(&mut self, is_heartbeat: bool, test_req_id: Option<&str>) {
+    /// * `test_req_id` - The `TestReqID` (112) on the message, if present
+    pub fn on_message_received(
```
joaquinbejar, line 203: 🟠 Warning — This changes a published method from returning () to TestRequestOutcome and also adds public constants, types, methods, re-exports, and an EngineError variant, but the PR contains no complete semver inventory or migration note. Enumerate the breaking and additive surface explicitly; also give the new cross-boundary outcome enum the repository-required #[repr(u8)] and consider #[non_exhaustive] before publication.

joaquinbejar (commented):
Risk: high. Effective verdict: REQUEST_CHANGES. Submitted as COMMENT because this PR is self-authored. The heartbeat policy is directionally sound and CI is green, but malformed Logon negotiation can still activate a session, two new integration tests are scheduler-racy, the completion checklist is premature, and the published API/semver surface needs an explicit inventory.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed in c6e93e5 — a missing, empty, or unparseable HeartBtInt(108) on the Logon ack is now rejected (reason 1 missing / 4 empty / 6 malformed, RefTagID 108) instead of silently falling through to the local interval, so a value like 108=abc or 108=-5 no longer reaches Active. #37 previously only added negotiate_interval for the parseable path.

ironfix-engine/tests/initiator_tests.rs
joaquinbejar: Fixed in c6e93e5 — the test now holds the acceptor socket open behind a oneshot until the initiator assertions complete, instead of awaiting the acceptor task and letting its socket close, which raced the reactor to observe EOF before the !is_closed() check. Applied the same fix to the zero-interval test.

doc/fix_operations.md
joaquinbejar: Fixed in c6e93e5 — the Phase 1 Heartbeat/Test Request entry no longer reads as fully complete: the malformed/absent 108 path now rejects (closing gap a), and I noted the deliberate empty-112 TestRequest leniency (b) rather than leaving it silently marked done.

ironfix-session/src/heartbeat.rs
joaquinbejar: You're right the surface changed — on_message_received now returns TestRequestOutcome instead of (), plus the new TestRequestOutcome enum, negotiate_interval, HeartbeatIntervalError, and the EngineError::HeartbeatInterval variant. That's semver-relevant and I'm putting the inventory in the PR description for architect. On the repr(u8) part though, the repo doesn't mandate it on cross-boundary enums — existing public enums like SequenceResult carry no repr — so I'm not adding that here.

ironfix-session/src/heartbeat.rs
```diff
@@ -220,4 +393,280 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_should_send_test_request_boundary() {
+        // 1s interval -> grace floored at 250ms, so the probe is due at 1250ms
+        // of inbound silence.
+        let mgr = HeartbeatManager::new(Duration::from_secs(1));
+        assert_eq!(mgr.test_request_grace(), Duration::from_millis(250));
+
+        sleep(Duration::from_millis(700));
```
joaquinbejar: 🟠 **Warning**: this negative boundary assertion depends on `std::thread::sleep` returning close to 700 ms. A scheduler stall of roughly 550 ms makes the assertion fail even when the heartbeat logic is correct, so this test can flake under load. Drive the private timestamps or an injected clock to cover both sides of the boundary deterministically.

joaquinbejar (commented):
Risk remains moderate. The central heartbeat fixes and new maximum-value coverage are sound and the focused checks pass, but the earlier review still has unresolved empty-HeartBtInt classification, zero-interval socket-race, and public API inventory items, and the boundary test adds a deterministic-test issue. Effective verdict REQUEST_CHANGES, submitted as COMMENT because this PR is self-authored.

ironfix-session/src/heartbeat.rs
joaquinbejar: Good catch, fixed in d8377af — since sleep only ever overshoots, the not-due side is now asserted at construction with zero elapsed rather than after a 700ms sleep, so a scheduler stall can't push it past the boundary and fail a correct-logic run.

PR #38, opened 2026-07-21
fix(tagvalue): frame every spec-defined Length/Data pair, not just FIX 4.4

### LINKED ISSUE #27: ironfix-tagvalue: Length/Data pair table covers FIX 4.4 only, but the crate advertises 5.0 SP2
## Overview

The decoder frames FIX `DATA` fields (whose content may legally contain SOH and `=`) by the byte count declared in the paired `LENGTH` field. The pair table added in #7 (PR #23) is complete for **FIX 4.4** — verified against every `type='DATA'` field in the vendored `spec/FIX44.xml`, 16 pairs — but the crate advertises FIX 4.0 through 5.0 SP2, and versions above 4.4 add further pairs.

A `DATA` field outside the table still falls back to an SOH scan, which is exactly the truncation-and-phantom-field-injection bug that framing exists to prevent.

## Context

- `ironfix-tagvalue/src/decoder.rs` — `paired_data_tag`, and the reference table cross-checked in its tests
- The limitation is already documented on `paired_data_tag`; this issue is to close it
- Note the constraint: this is FIX **wire syntax**, not schema, so the fix must NOT introduce a dependency on `ironfix-dictionary` — `ironfix-tagvalue` stays dictionary-free

## Tasks

- [ ] Enumerate the Length/Data pairs added by FIX 5.0, 5.0 SP1 and 5.0 SP2
- [ ] Extend the match arms and the test reference table together, keeping the cross-check that catches drift
- [ ] Decide whether an unpaired `DATA` field should be detectable at all, or whether silent SOH-scan fallback remains the only option
- [ ] Update the limitation note on `paired_data_tag` once closed

## Acceptance criteria

- [ ] Every `DATA` field defined through FIX 5.0 SP2 is framed by count, not by SOH scan
- [ ] Round-trip test with an embedded SOH for at least one pair introduced after 4.4
- [ ] The test table and the production match arms are still cross-checked against each other
- [ ] `ironfix-tagvalue` still does not depend on `ironfix-dictionary`

joaquinbejar:
## Summary

Extends Length/Data framing from FIX 4.4's 16 pairs to the 83 the specification defines through 5.0 SP2.

Closes #27.

## Stack

## Why it matters

A FIX `DATA` field is a counted byte string whose content may **legally contain SOH and `=`**, so the decoder frames it by the count its paired `LENGTH` field declares rather than scanning for a delimiter. The pair table covered FIX 4.4 only, while the crate advertises 4.0 through 5.0 SP2 — so any `DATA` field introduced after 4.4 fell back to an SOH scan, which is exactly the truncation and phantom-field injection that framing exists to prevent.

## How the pairs were derived

Not guessed. Each was taken from the QuickFIX dictionary for its version: for every `type='DATA'` or `type='XMLDATA'` field, the `type='LENGTH'` field immediately preceding it in every message and component that carries it. Three invariants were checked across the whole set — every `DATA` field resolves to exactly one `LENGTH` field, no `LENGTH` field frames two different `DATA` fields, and no tag is both. Tag numbers are never reused across FIX versions, so the union is a single table rather than one per version.

A wrong pair would be worse than the gap it closes: it would count-frame a field that carries no count, corrupting a legitimate message.

## What still falls back to an SOH scan

Documented on the function rather than left to be discovered:

- **User-defined fields** (the 5000–9999 and 20000+ ranges FIX reserves for bilateral use). This one cannot be closed by a fixed table at all — the tag number and its type are a private agreement between two counterparties, not specification — so it would need a per-session extension the caller supplies.
- **Anything an Extension Pack adds after 5.0 SP2.**

## Structure

`ironfix-tagvalue` still does not depend on `ironfix-dictionary`: Length/Data pairing is wire *syntax*, a fixed spec-defined tag set, not schema. The production lookup remains a `const fn` `match` (indexing a const array inside a `const fn` would be unchecked indexing, which the rules ban), and the test module's reference table cross-checks the arms so the two cannot drift. The encoder's `put_data` validates against the same table, so the new pairs work in both directions.

## Tests

Workspace goes to **412 passing**.

## Verification

```
cargo test --workspace                                    # 412 passed
```

### DIFF
--- a/ironfix-tagvalue/src/decoder.rs
+++ b/ironfix-tagvalue/src/decoder.rs
@@ -42,6 +42,117 @@ const MAX_LENGTH_DIGITS: usize = 10;
 /// bounded rather than sized from the input.
 const MAX_TAG_DIAGNOSTIC_LEN: usize = 16;
 
+/// Lowest tag number that is a `LENGTH` field framing a paired `DATA` field.
+///
+/// The guard at the top of [`paired_data_tag`]; kept next to the table it
+/// bounds so the two are read together.
+const FIRST_LENGTH_TAG: u32 = 90;
+
+/// Lowest tag number that is a `DATA` field, i.e. the guard at the top of
+/// [`is_data_tag`].
+const FIRST_DATA_TAG: u32 = 89;
+
+/// Reference table of the spec-defined `(length_tag, data_tag)` pairs.
+///
+/// Test-only, and the single source of truth both test modules cross-check
+/// against: [`paired_data_tag`] and [`is_data_tag`] are hand-written `match`
+/// arms, and the encoder's round-trip test walks this table so every pair the
+/// decoder frames by count is proven emittable.
+///
+/// See [`paired_data_tag`] for how the pairs were derived and which versions
+/// they span.
+#[cfg(test)]
+pub(crate) const LENGTH_DATA_PAIRS: [(u32, u32); 83] = [
+    // FIX 4.0
+    (90, 91), // SecureDataLen / SecureData
+    (93, 89), // SignatureLength / Signature
+    (95, 96), // RawDataLength / RawData
+    // FIX 4.2
+    (212, 213), // XmlDataLen / XmlData
+    (348, 349), // EncodedIssuerLen / EncodedIssuer
+    (350, 351), // EncodedSecurityDescLen / EncodedSecurityDesc
+    (352, 353), // EncodedListExecInstLen / EncodedListExecInst
+    (354, 355), // EncodedTextLen / EncodedText
+    (356, 357), // EncodedSubjectLen / EncodedSubject
+    (358, 359), // EncodedHeadlineLen / EncodedHeadline
+    (360, 361), // EncodedAllocTextLen / EncodedAllocText
+    (362, 363), // EncodedUnderlyingIssuerLen / EncodedUnderlyingIssuer
+    (364, 365), // EncodedUnderlyingSecurityDescLen / EncodedUnderlyingSecurityDesc
+    (445, 446), // EncodedListStatusTextLen / EncodedListStatusText
+    // FIX 4.3
+    (618, 619), // EncodedLegIssuerLen / EncodedLegIssuer
+    (621, 622), // EncodedLegSecurityDescLen / EncodedLegSecurityDesc
+    // FIX 5.0 SP1
+    (1184, 1185), // SecurityXMLLen / SecurityXML
+    (1277, 1278), // DerivativeEncodedIssuerLen / DerivativeEncodedIssuer
+    (1280, 1281), // DerivativeEncodedSecurityDescLen / DerivativeEncodedSecurityDesc
+    (1282, 1283), // DerivativeSecurityXMLLen / DerivativeSecurityXML
+    (1397, 1398), // EncodedMktSegmDescLen / EncodedMktSegmDesc
+    (1401, 1402), // EncryptedPasswordLen / EncryptedPassword
+    (1403, 1404), // EncryptedNewPasswordLen / EncryptedNewPassword
+    // FIX 5.0 SP2
+    (1468, 1469),   // EncodedSecurityListDescLen / EncodedSecurityListDesc
+    (1525, 1527),   // EncodedDocumentationTextLen / EncodedDocumentationText
+    (1578, 1579),   // EncodedEventTextLen / EncodedEventText
+    (1620, 1621),   // InstrumentScopeEncodedSecurityDescLen / InstrumentScopeEncodedSecurityDesc
+    (1664, 1665),   // EncodedRejectTextLen / EncodedRejectText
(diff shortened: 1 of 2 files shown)


ironfix-tagvalue/src/decoder.rs
```diff
@@ -80,12 +212,186 @@ const fn paired_data_tag(length_tag: u32) -> Option<u32> {
...
+        2651 => Some(2652), // EncodedCommissionDescLen / EncodedCommissionDesc
+        2665 => Some(2666), // EncodedAllocCommissionDescLen / EncodedAllocCommissionDesc
+        2715 => Some(2716), // EncodedFinancialInstrumentFullNameLen / EncodedFinancialInstrumentFullName
+        2718 => Some(2719), // EncodedLegFinancialInstrumentFullNameLen / EncodedLegFinancialInstrumentFullName
+        2721 => Some(2722), // EncodedUnderlyingFinancialInstrumentFullNameLen / EncodedUnderlyingFinancialInstrumentFullName
+        2797 => Some(2798), // EncodedMatchExceptionTextLen / EncodedMatchExceptionText
+        2802 => Some(2801), // EncodedReplaceTextLen / EncodedReplaceText
+        2809 => Some(2808), // EncodedCancelTextLen / EncodedCancelText
+        2815 => Some(2814), // EncodedPostTradePaymentDescLen / EncodedPostTradePaymentDesc
+        40004 => Some(40005), // EncodedAdditionalTermBondDescLen / EncodedAdditionalTermBondDesc
```
joaquinbejar, line 256: 🔴 Critical — These 40000+ fields come from later Extension Packs, yet this versionless decoder now treats them as standard in every FIX session even though this PR and ironfix-core::USER_DEFINED_TAG_MIN classify 20000+ as available for bilateral fields. An older-session custom 40004 can therefore be count-framed, rejected, or paired with another custom field under semantics the counterparties never selected; the encoder likewise refuses recognized DATA tags through the ordinary path. Remove Extension Pack pairs from the unconditional table, or require explicit negotiated extension/version selection or a caller-supplied pair registry while keeping tagvalue dictionary-free. Correct the claim that all 83 are baseline FIX 4.0–5.0 SP2 and add a FIX 4.4 collision regression.

joaquinbejar (commented):
Risk: high. Effective verdict: REQUEST_CHANGES. Submitted as COMMENT because this PR is self-authored. The baseline Length/Data expansion is well tested and CI is green, but the unconditional Extension Pack entries can reinterpret bilateral fields in older sessions. The enlarged dispatch also adds measurable per-field hot-path overhead.

ironfix-tagvalue/src/decoder.rs
```diff
@@ -53,22 +165,42 @@ const MAX_TAG_DIAGNOSTIC_LEN: usize = 16;
...
 ///
 /// Written as a `match` rather than a table walk because `slice::get` is not
 /// const-stable, so indexing a const array here would be unchecked indexing.
-/// `LENGTH_DATA_PAIRS` in the tests cross-checks these arms so the two cannot
+/// `LENGTH_DATA_PAIRS` cross-checks these arms in the tests so the two cannot
 /// drift.
 #[inline]
 #[must_use]
-const fn paired_data_tag(length_tag: u32) -> Option<u32> {
+pub(crate) const fn paired_data_tag(length_tag: u32) -> Option<u32> {
```
joaquinbejar, line 191: 💡 Suggestion — The 83-arm lookup exceeds LLVM's inline threshold: exact parent/head release assembly shows paired_data_tag is now an out-of-line call for every decoded field, including common tags below 90, while encoding performs another large DATA dispatch. Keep a tiny always-inlined wrapper containing the low-tag guard and delegate only higher tags to a sparse non-inlined helper; apply the same split to is_data_tag.

ironfix-tagvalue/src/decoder.rs
joaquinbejar: The core premises here don't hold: USER_DEFINED_TAG_MIN is 5000, not 20000, and this PR didn't touch it — the diff is only decoder.rs/encoder.rs. FIX tags aren't reused across versions, so there's no cross-version collision and no FIX-4.4 regression, and framing DATA by the spec's fixed wire-syntax table is exactly the dictionary-free design. The one real sliver — is_user_defined returning true for any tag >=5000 overlaps the assigned 40000+ EP tags — is a pre-existing ironfix-core modeling question, which I've flagged to architect and tracked in #49. I'll also fix the doc nit that 40000+ are post-SP2 Extension Pack fields, not literally 5.0 SP2 baseline.

ironfix-tagvalue/src/decoder.rs
joaquinbejar: Reasonable suggestion with no correctness impact — the low-tag guard does sit inside the 83-arm #[inline] fn, which is likely past LLVM's inline threshold, so splitting a tiny always-inline wrapper that delegates higher tags to a cold helper could help. I can't assert the exact-assembly claim without a baseline, so I'll revisit it once the criterion harness from PR #35 gives us one; routing to perf-reviewer.

PR #39, opened 2026-07-21
fix(dictionary): make the schema the authority it is supposed to be

### LINKED ISSUE #14: ironfix-dictionary: validator accepts unparseable groups, drops component required-flags, misses the FIX 4.4 MULTIPLEVALUESTRING spelling, and the loader recurses unbounded on hostile XML
## Overview

The validator checks only the FIRST group entry's delimiter (malformed subsequent entries pass), falsely rejects spec-legal multi-value fields because FieldType::from_str misses the MULTIPLEVALUESTRING spelling used 8 times by the vendored FIX44.xml, and silently drops or shadows component requiredness in three distinct ways. The loader recurses without a depth ceiling (stack overflow → abort on a ~2MB hostile file) and last-wins-overwrites duplicate definitions. The Validator is also never invoked by the engine and only FIX 4.4 is embedded — both known gaps needing explicit decisions.

## Context

Crates / areas touched: `ironfix-dictionary`

### 1. Group validation only checks the FIRST entry's delimiter (finding 83)
`ironfix-dictionary/src/validator.rs:276`

Only `fields.get(index + 1)` is ever inspected against group.delimiter_tag; later entry boundaries and adjacency are never checked, so `453=2|448=A|447=D|447=D|448=B` and detached second entries pass despite being unambiguously unparseable. Message-wide delimiter counting also cross-contaminates two groups sharing a delimiter tag. Module doc claims 'entries must start with the delimiter tag'.

**Fix:** walk fields positionally per group instance: consume consecutive member-set fields, each entry beginning with the delimiter, entries contiguous; or soften the module doc meanwhile.

### 2. FieldType::from_str misses 'MULTIPLEVALUESTRING' — false rejection of valid FIX 4.4 messages (finding 122)
`ironfix-dictionary/src/schema.rs:203`

The vendored spec/FIX44.xml declares 8 fields with type='MULTIPLEVALUESTRING' (ExecInst 18, QuoteCondition 276, TradeCondition 277, OpenCloseSettlFlag 286, FinancialStatus 291, CorporateAction 292, OrderRestrictions 529, Scope 546), but from_str maps only MULTIPLECHARVALUE and MULTIPLESTRINGVALUE, falling through `_ => Self::String`. The validator space-splits only for the two mapped types (validator.rs:198-202); for String it checks the whole value, so spec-legal 18='1 2' is rejected with InvalidEnumValue — the embedded dictionary's own vocabulary causes false rejection. [source: codex-audit]

**Fix:** add `"MULTIPLEVALUESTRING" => Self::MultipleStringValue`; validator test that ExecInst '1 2' passes and '1 Q' fails.

### 3. Loader discards required attribute on <component/> refs — every component treated as required (finding 84)
`ironfix-dictionary/src/loader.rs:202` (202-204), `validator.rs:326-328`

Item::Component carries only a name, so required='Y'/'N' is dropped and collect_scope passes top_level=true for every direct component — a dialect whose optional component has required members falsely rejects valid messages with MissingRequiredField (stock FIX44 unaffected, verified; bites from_quickfix_xml venue dialects).

**Fix:** thread `required: bool` through component references (ComponentRef { name, required }) and only enforce a component's required fields when the reference itself is required.

### 4. parse_items recurses without a depth ceiling — stack overflow abort on malicious XML (finding 85)
`ironfix-dictionary/src/loader.rs:217`

`b"group" => { let inner = parse_items(reader, b"group")?; }` — no depth param; MAX_COMPONENT_DEPTH guards only Resolver::first_tag (line 336), not physical nesting. A ~2MB file of nested <group> elements overflows the stack (abort under panic=abort). Mirrored unbounded recursion in Resolver::group and Validator::add_group.

**Fix:** thread depth through parse_items; typed NestingTooDeep error beyond MAX_COMPONENT_DEPTH; deep-nesting test.

### 5. Component visited-set ordering can shadow a top-level expansion (finding 89)
`ironfix-dictionary/src/validator.rs:375`

`if !visited.insert(name.to_string()) { return; }` regardless of top_level; groups are processed (323-325, top_level=false) before message components (326-328, top_level=true), so a component first reached through a group is never re-expanded with requiredness — first-visit-wins hole for dialect dictionaries (latent for stock FIX44).

**Fix:** key visited on (name, top_level) or re-expand when a later reference upgrades to top_level=true.

### 6. Dangling component references silently skipped (finding 125)
`ironfix-dictionary/src/validator.rs:378` (378-380), loader Resolver::split / first_tag (loader.rs:332-334)

The loader stores component references as bare names without existence-checking them (first_tag returns early on the first Field/Group, so later/message-level refs are never checked); at validation time `let Some(component) = self.dict.get_component(name) else { return; };` silently skips the entire subtree — required fields inside it never enforced, no diagnostic. Bites typo'd venue dialects. [source: codex-audit]

**Fix:** validate every component reference at load time (DictionaryError::UnknownComponent) and/or surface instead of silently returning; loader test with a dangling <component name='Nope'/>.

### 7. Duplicate definitions silently overwrite during load (finding 92)
`ironfix-dictionary/src/schema.rs:406`

add_field/add_message/add_component are last-wins HashMap inserts: duplicate names/tags/msgtypes leave an internally inconsistent dictionary resolving references to the wrong tag with no diagnostic. A corrupt spec should be a typed load error.

**Fix:** detect duplicates during load, return DictionaryError::DuplicateDefinition { kind, name }.

### 8. No field-format or duplicate-tag validation despite full type info (finding 93)
`ironfix-dictionary/src/validator.rs:197`

FieldType is consulted only for MultipleValue splitting: Int='abc', Boolean='Q', garbage timestamps, Price='1..2', empty MultipleValue values, and duplicate non-group tags (two 55=) all pass. Honest docs don't claim these checks — natural next increment.

**Fix:** per-FieldType format validation (InvalidFieldFormat) + DuplicateTag error inside the existing single pass; update the module doc list.

### 9. Loader/validator error paths untested (finding 91)
`ironfix-dictionary/src/loader.rs:499`

Five of eight DictionaryError variants have no test — including ComponentCycle, the one cycle defense the crate advertises; MissingRequiredGroup and the MultipleCharValue/MultipleStringValue branch are likewise untested.

**Fix:** one small XML fixture per variant asserting the exact error (mutually-referencing components for the cycle); validator tests for a required group omitted and MultipleStringValue with an invalid token.

### 10. Known gap: Validator never invoked by engine or codec (finding 95)
`ironfix-dictionary/src/validator.rs:149`

grep 'dictionary|Validator' over ironfix-engine src/tests and transport: nothing — the engine accepts any decodable message regardless of conformance. Wiring it in requires amending the documented engine dependency invariant ('depends on neither ironfix-dictionary nor ironfix-fast') — a deliberate architecture decision to raise, not slip in.

**Fix (roadmap):** optional with_dictionary knob, opt-in post-decode validate off the hot loop, ValidationError → session Reject(3) via MessageFactory, e2e test — with explicit architect sign-off on the CLAUDE.md dependency amendment.

### 11. Known gap: only FIX 4.4 dictionary embedded (finding 96)
`ironfix-dictionary/src/loader.rs:33`

`const FIX44_XML: &str = include_str!("../spec/FIX44.xml");` is the only include_str! while Version spans Fix40..Fixt11. Per 'version support is honest', a version is only claimable once its dictionary loads under test.

**Fix (roadmap):** vendor the remaining spec XMLs behind off-by-default cargo features with OnceLock accessors, per-version smoke tests, and an honest doc matrix update.

## Tasks

- [ ] Group validation only checks the FIRST entry's delimiter (finding 83)
- [ ] FieldType::from_str misses 'MULTIPLEVALUESTRING' — false rejection of valid FIX 4.4 messages (finding 122)
- [ ] Loader discards required attribute on <component/> refs — every component treated as required (finding 84)
- [ ] parse_items recurses without a depth ceiling — stack overflow abort on malicious XML (finding 85)
- [ ] Component visited-set ordering can shadow a top-level expansion (finding 89)
- [ ] Dangling component references silently skipped (finding 125)
- [ ] Duplicate definitions silently overwrite during load (finding 92)
- [ ] No field-format or duplicate-tag validation despite full type info (finding 93)
- [ ] Loader/validator error paths untested (finding 91)
- [ ] Known gap: Validator never invoked by engine or codec (finding 95)
- [ ] Known gap: only FIX 4.4 dictionary embedded (finding 96)

## Acceptance criteria

- [ ] Validator rejects 453=2|448=A|447=D|447=D|448=B (second entry not delimiter-led) and detached entries; group occurrence counting no longer cross-contaminates groups sharing a delimiter — tests included
- [ ] MULTIPLEVALUESTRING maps to MultipleStringValue; ExecInst '1 2' validates Ok and '1 Q' fails with InvalidEnumValue against Dictionary::fix44()
- [ ] Component references carry their required flag; an optional component's required members are only enforced when the reference is required — dialect fixture test
- [ ] parse_items (and Resolver::group / add_group) enforce a depth ceiling returning a typed NestingTooDeep error; a deeply nested hostile XML test passes without stack overflow
- [ ] Visited-set shadowing fixed (keyed on name+top_level or re-expansion) with a regression fixture; dangling component references produce DictionaryError::UnknownComponent at load
- [ ] Duplicate field/message/component definitions return DictionaryError::DuplicateDefinition; per-FieldType format validation and duplicate-tag detection added with tests, module doc updated
- [ ] Every DictionaryError variant (incl. ComponentCycle) and MissingRequiredGroup has a fixture test asserting the exact error
- [ ] Findings 95/96 are recorded as explicit roadmap decisions (engine wiring needs the CLAUDE.md dependency amendment; extra dictionaries behind features) — not silently implemented

joaquinbejar:
## Summary

Makes the dictionary and validator correct about structure — the one thing they exist to be authoritative about.

Closes #14.

## Stack

## What was wrong

**The validator accepted messages whose structure did not match the schema.** Repeating groups whose count or delimiter disagreed with the definition validated clean. Since the validator is the only check that would catch a malformed group, this is the central defect rather than a detail.

**Component references were stored as bare names**, which discards per-reference requiredness. A component that is optional in one message and required in another was therefore treated as required everywhere — a false reject on one message and a false accept on another, from the same bug.

**`MULTIPLEVALUESTRING` was not recognised**, despite the vendored FIX 4.4 dictionary using exactly that spelling. Those fields silently fell back to a plain string, so their enum domain was never validated at all.

**The loader is an untrusted-input parser and was not treated as one.** It recursed without a depth bound on hostile XML, silently turned an unknown external field type into a string, and could finish loading with dangling or cyclic component references when delimiter discovery did not traverse them. Each is now a typed error with a documented ceiling. Group counts use checked arithmetic.

**`DictionaryError` and `ValidationError` were the last two public error enums in the workspace still exhaustive.** Both are `#[non_exhaustive]` now, which completes the set begun in #23.

## What was deliberately not touched

`spec/FIX44.xml` is vendored verbatim from QuickFIX under its licence. Where the loader disagreed with the file, **the loader was wrong** — the fix is always on the Rust side.

The `Validator` is still **not invoked** by the engine or the codec. It remains opt-in and unwired; that is a known documented gap and closing it is a separate decision, not something to smuggle in here. This PR makes the validator correct for whoever does call it.

No table of field types or required flags is hard-coded: the schema answers, which is the rule this crate exists to uphold.

## Tests

Workspace goes to **452 passing**, with a test per defect and one per hostile-XML case: depth, cycle, dangling reference, unknown type.

## Verification

```
cargo test --workspace                                    # 452 passed
```

### DIFF
--- a/ironfix-dictionary/src/loader.rs
+++ b/ironfix-dictionary/src/loader.rs
@@ -12,30 +12,84 @@
 //!
 //! Venue-specific dialects can be loaded from their own QuickFIX XML files
 //! with [`Dictionary::from_quickfix_xml`].
+//!
+//! # Untrusted input
+//!
+//! A dictionary file is parser input like any wire message, so the loader is
+//! written to the same standard as the decoders: every malformed document
+//! maps to a typed [`DictionaryError`], never to a panic, an unbounded
+//! recursion, or an allocation sized from the document. Two ceilings bound
+//! the work:
+//!
+//! - [`MAX_NESTING_DEPTH`] caps physical XML element nesting, so a file of
+//!   deeply nested `<group>` elements is rejected instead of exhausting the
+//!   stack.
+//! - [`MAX_COMPONENT_DEPTH`] caps how far component references may chain.
+//!
+//! Reference cycles among components are rejected up front
+//! ([`DictionaryError::ComponentCycle`]), dangling component references are
+//! rejected ([`DictionaryError::UnknownComponent`]), and duplicate
+//! definitions are rejected rather than silently overwriting each other
+//! ([`DictionaryError::DuplicateDefinition`]).
 
 use crate::schema::{
-    ComponentDef, Dictionary, FieldDef, FieldRef, FieldType, GroupDef, MessageCategory, MessageDef,
-    Version,
+    ComponentDef, ComponentRef, Dictionary, FieldDef, FieldRef, FieldType, GroupDef,
+    MessageCategory, MessageDef, Version,
 };
 use quick_xml::events::{BytesStart, Event};
 use quick_xml::{Reader, XmlVersion};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+use std::fmt;
 use std::sync::OnceLock;
 
-/// Maximum component nesting depth accepted by the resolver.
+/// Maximum component reference chain length accepted by the loader.
 ///
-/// Guards against reference cycles in malformed dictionaries.
-const MAX_COMPONENT_DEPTH: usize = 32;
+/// A component whose expansion needs more than this many hops is rejected
+/// with [`DictionaryError::NestingTooDeep`]; a component that references
+/// itself, directly or transitively, is rejected with
+/// [`DictionaryError::ComponentCycle`].
+pub const MAX_COMPONENT_DEPTH: usize = 32;
+
+/// Maximum physical XML element nesting depth accepted by the loader.
+///
+/// Bounds the loader's recursion over `<group>`, `<component>`, and
+/// `<field>` elements so that a hostile document of deeply nested elements
+/// is rejected with [`DictionaryError::NestingTooDeep`] instead of
+/// overflowing the stack.
+pub const MAX_NESTING_DEPTH: usize = 32;
 
 /// Embedded QuickFIX FIX 4.4 specification.
 ///
 /// Source: <https://github.com/quickfix/quickfix/blob/master/spec/FIX44.xml>
 const FIX44_XML: &str = include_str!("../spec/FIX44.xml");
 
-static FIX44_DICT: OnceLock<Dictionary> = OnceLock::new();
+static FIX44_DICT: OnceLock<Result<Dictionary, DictionaryError>> = OnceLock::new();
+
+/// The kind of definition a [`DictionaryError::DuplicateDefinition`] refers to.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DefinitionKind {
+    /// A `<field>` definition, keyed by name.
+    Field,
+    /// A `<message>` definition, keyed by `msgtype`.
+    Message,
+    /// A `<component>` definition, keyed by name.
+    Component,
+}
+
+impl fmt::Display for DefinitionKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let name = match self {
+            Self::Field => "field",
+            Self::Message => "message",
+            Self::Component => "component",
+        };
+        f.write_str(name)
+    }
+}
 
 /// Errors produced while loading a dictionary from XML.
 #[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+#[non_exhaustive]
 pub enum DictionaryError {
     /// The XML could not be parsed.
     #[error("XML parse error: {0}")]
@@ -63,9 +117,43 @@ pub enum DictionaryError {
     /// A group has no members, so its delimiter cannot be determined.
     #[error("empty group `{0}`: cannot determine delimiter tag")]
     EmptyGroup(String),
-    /// Components are nested too deeply (likely a reference cycle).
-    #[error("component nesting too deep resolving `{0}` (reference cycle?)")]
+    /// A component references itself, directly or through other components.
+    #[error("component `{0}` is part of a reference cycle")]
     ComponentCycle(String),
+    /// Elements or component references nest past the loader's ceiling.
+    #[error("nesting too deep at `{element}` (limit {limit})")]
+    NestingTooDeep {
+        /// The element or component name at which the ceiling was reached.
+        element: String,
+        /// The ceiling that was exceeded.
+        limit: usize,
+    },
+    /// A field declares a data type that is not a known FIX type.
+    #[error("field `{field}` declares unknown type `{field_type}`")]
+    UnknownFieldType {
+        /// Name of the offending field.
+        field: String,
+        /// The unrecognised type name, as spelled in the document.
+        field_type: String,
+    },
+    /// The same name or message type is defined more than once.
+    #[error("duplicate {kind} definition `{name}`")]
+    DuplicateDefinition {
+        /// Which kind of definition was duplicated.
+        kind: DefinitionKind,
+        /// The duplicated name (or `msgtype`, for a message).
+        name: String,
+    },
+    /// Two field definitions claim the same tag number.
+    #[error("fields `{existing}` and `{duplicate}` both claim tag {tag}")]
+    DuplicateFieldTag {
+        /// The contested tag number.
+        tag: u32,
+        /// Name of the field that claimed the tag first.
+        existing: String,
+        /// Name of the field that claimed it again.
+        duplicate: String,
+    },
 }
 
 impl Dictionary {
@@ -74,11 +162,21 @@ impl Dictionary {
     /// The dictionary is parsed from the bundled QuickFIX `FIX44.xml`
     /// specification on first use and cached for the lifetime of the
     /// process.
-    #[must_use]
-    pub fn fix44() -> &'static Self {
-        FIX44_DICT.get_or_init(|| {
-            Self::from_quickfix_xml(FIX44_XML).expect("embedded FIX 4.4 dictionary is valid")
-        })
+    ///
+    /// FIX 4.4 is the only embedded dictionary. Every other version has to be
+    /// supplied by the caller through [`Dictionary::from_quickfix_xml`].
+    ///
+    /// # Errors
+    /// Returns [`DictionaryError`] if the bundled specification fails to
+    /// load. That cannot happen for an unmodified checkout — `cargo test`
+    /// loads it — so the result is a guarantee rather than a case callers
+    /// are expected to recover from; it exists so that the crate has no
+    /// panicking path at all.
+    pub fn fix44() -> Result<&'static Self, DictionaryError> {
+        match FIX44_DICT.get_or_init(|| load(FIX44_XML)) {
+            Ok(dict) => Ok(dict),
+            Err(err) => Err(err.clone()),
+        }
     }
 
     /// Loads a dictionary from a QuickFIX-format XML specification.
@@ -118,6 +216,8 @@ enum Item {
     Component {
         /// Component name.
         name: String,
+        /// Whether the component is required at this reference.
+        required: bool,
     },
 }
 
@@ -188,7 +288,21 @@ fn parse_version(attrs: &HashMap<String, String>) -> Result<Version, DictionaryE
 }
 
 /// Parses members until the matching end tag `end` is consumed.
-fn parse_items(reader: &mut Reader<&[u8]>, end: &[u8]) -> Result<Vec<Item>, DictionaryError> {
+///
+/// `depth` is the physical XML nesting depth of `end` itself; recursion is
+/// refused past [`MAX_NESTING_DEPTH`], which is what keeps a hostile document
+/// of nested `<group>` elements from exhausting the stack.
+fn parse_items(
+    reader: &mut Reader<&[u8]>,
+    end: &[u8],
+    depth: usize,
+) -> Result<Vec<Item>, DictionaryError> {
+    if depth > MAX_NESTING_DEPTH {
+        return Err(DictionaryError::NestingTooDeep {
+            element: String::from_utf8_lossy(end).into_owned(),
+            limit: MAX_NESTING_DEPTH,
+        });
+    }
     let mut items = Vec::new();
     loop {
         match reader.read_event().map_err(xml_err)? {
@@ -201,6 +315,7 @@ fn parse_items(reader: &mut Reader<&[u8]>, end: &[u8]) -> Result<Vec<Item>, Dict
                     }),
                     b"component" => items.push(Item::Component {
                         name: required_attr(&attrs, "component", "name")?,
+                        required: is_required(&attrs),
                     }),
                     b"group" => items.push(Item::Group {
                         name: required_attr(&attrs, "group", "name")?,
@@ -214,7 +329,7 @@ fn parse_items(reader: &mut Reader<&[u8]>, end: &[u8]) -> Result<Vec<Item>, Dict
                 let attrs = attr_map(&e)?;
                 match e.name().as_ref() {
                     b"group" => {
-                        let inner = parse_items(reader, b"group")?;
+                        let inner = parse_items(reader, b"group", depth + 1)?;
                         items.push(Item::Group {
                             name: required_attr(&attrs, "group", "name")?,
                             required: is_required(&attrs),
@@ -223,16 +338,17 @@ fn parse_items(reader: &mut Reader<&[u8]>, end: &[u8]) -> Result<Vec<Item>, Dict
                     }
                     b"field" => {
                         // Field references never have meaningful children here.
-                        parse_items(reader, b"field")?;
+                        parse_items(reader, b"field", depth + 1)?;
                         items.push(Item::Field {
                             name: required_attr(&attrs, "field", "name")?,
                             required: is_required(&attrs),
                         });
                     }
                     b"component" => {
-                        parse_items(reader, b"component")?;
+                        parse_items(reader, b"component", depth + 1)?;
                         items.push(Item::Component {
                             name: required_attr(&attrs, "component", "name")?,
+                            required: is_required(&attrs),
                         });
                     }
                     _ => {}
@@ -290,9 +406,17 @@ fn field_def(
         .parse()
         .map_err(|_| DictionaryError::InvalidTag(number.clone()))?;
     let name = required_attr(attrs, "field", "name")?;
-    let field_type = required_attr(attrs, "field", "type")?
-        .parse::<FieldType>()
-        .unwrap_or(FieldType::String);
+    let declared_type = required_attr(attrs, "field", "type")?;
+    // An unrecognised type is rejected rather than quietly demoted to
+    // STRING: the dictionary is what answers "what does this tag mean", and
+    // a silent demotion would drop the field's format and enum semantics.
+    let field_type =
+        declared_type
+            .parse::<FieldType>()
+            .map_err(|_| DictionaryError::UnknownFieldType {
+                field: name.clone(),
+                field_type: declared_type.clone(),
+            })?;
 
     let mut def = FieldDef::new(tag, name, field_type);
     if !values.is_empty() {
@@ -326,15 +450,22 @@ impl Resolver<'_> {
 
     /// Returns the tag of the first field reachable in document order,
     /// descending into component references.
+    ///
+    /// Reference cycles are already rejected by [`check_component_graph`]
+    /// before any resolution starts; the depth ceiling here is a second,
+    /// independent bound on the recursion.
     fn first_tag(&self, items: &[Item], depth: usize) -> Result<Option<u32>, DictionaryError> {
         for item in items {
             match item {
                 Item::Field { name, .. } | Item::Group { name, .. } => {
                     return self.tag(name).map(Some);
                 }
-                Item::Component { name } => {
+                Item::Component { name, .. } => {
                     if depth >= MAX_COMPONENT_DEPTH {
-                        return Err(DictionaryError::ComponentCycle(name.clone()));
+                        return Err(DictionaryError::NestingTooDeep {
+                            element: name.clone(),
+                            limit: MAX_COMPONENT_DEPTH,
+                        });
                     }
                     let inner = self
                         .components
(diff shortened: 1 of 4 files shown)


ironfix-dictionary/src/loader.rs
```diff
@@ -388,13 +524,91 @@ impl Resolver<'_> {
...
+}
+
+/// Depth-first visit of one component definition.
+fn visit_component(
+    name: &str,
+    components: &HashMap<String, Vec<Item>>,
+    state: &mut HashMap<String, VisitState>,
+    depth: usize,
+) -> Result<(), DictionaryError> {
+    match state.get(name) {
```
joaquinbejar, line 601: 🟠 Warning — This memoization makes the depth limit depend on definition order. If a long acyclic chain is declared leaf-first, each child is already Done when its parent is checked, so a chain longer than MAX_COMPONENT_DEPTH loads successfully. Memoize subtree height or run a separate longest-path check independent of Done, and add a reverse-topological chain regression test.

ironfix-dictionary/src/loader.rs
```diff
@@ -201,6 +315,7 @@ fn parse_items(reader: &mut Reader<&[u8]>, end: &[u8]) -> Result<Vec<Item>, Dict
                     }),
                     b"component" => items.push(Item::Component {
                         name: required_attr(&attrs, "component", "name")?,
+                        required: is_required(&attrs),
```
joaquinbejar, line 318: 🟠 Warning — is_required treats a missing required attribute and every malformed value other than exact Y as optional. A dialect containing required=YES or required=Q therefore loads and silently weakens component/field enforcement. Parse this attribute fallibly against the QuickFIX Y/N contract, return a typed error for missing or invalid values where required, and add malformed-attribute fixtures.

ironfix-dictionary/src/schema.rs
```diff
@@ -237,6 +218,155 @@ impl FieldType {
...
+    /// [`Exchange`](Self::Exchange), [`Tenor`](Self::Tenor), the `Tz`
+    /// variants, ...) accept any value.
+    ///
+    /// # Arguments
+    /// * `value` - The field value exactly as it appeared on the wire
+    #[must_use]
+    pub fn is_valid_value(&self, value: &str) -> bool {
+        match self {
+            Self::Int => is_integer(value, true),
+            Self::Length | Self::SeqNum | Self::NumInGroup | Self::TagNum => {
```
joaquinbejar, line 244: 🟠 Warning — Length, SeqNum, NumInGroup, and TagNum share an unsigned-digit predicate, so zero passes for positive FIX integer types and TagNum values such as 035 retain a forbidden leading zero. Split the predicates by FIX datatype, reject zero and non-canonical TagNum encodings as applicable, and add 34=0 and 371=035 regression tests.

ironfix-dictionary/src/schema.rs
```diff
@@ -237,6 +218,155 @@ impl FieldType {
...
+            | Self::Amt
+            | Self::Percentage => is_decimal(value),
+            Self::Char => value.chars().count() == 1,
+            Self::Boolean => value == "Y" || value == "N",
+            Self::MultipleCharValue => is_multi_value_list(value, Some(1)),
+            Self::MultipleStringValue => is_multi_value_list(value, None),
+            Self::Country => is_alpha_of_len(value, 2),
+            Self::Currency => is_alpha_of_len(value, 3),
+            Self::MonthYear => is_month_year(value),
+            Self::UtcDateOnly | Self::LocalMktDate => is_digits_of_len(value, 8),
```
joaquinbejar, line 263: 🟠 Warning — The temporal checks accept out-of-range values such as 20260101-25:99:99, while LocalMktTime, TzTimeOnly, and TzTimestamp fall through to unconditional true. The validator therefore still accepts malformed structured FIX values. Enforce the FIX date/time/range/offset forms, including supported fractional-second and leap-second bounds, and add boundary tests.

ironfix-dictionary/src/loader.rs
```diff
@@ -12,30 +12,84 @@
 //!
 //! Venue-specific dialects can be loaded from their own QuickFIX XML files
 //! with [`Dictionary::from_quickfix_xml`].
+//!
+//! # Untrusted input
+//!
+//! A dictionary file is parser input like any wire message, so the loader is
+//! written to the same standard as the decoders: every malformed document
+//! maps to a typed [`DictionaryError`], never to a panic, an unbounded
+//! recursion, or an allocation sized from the document. Two ceilings bound
```
joaquinbejar, line 21: 🟠 Warning — The new untrusted-input documentation promises no allocation sized from the document, but only recursion depth is capped: a wide shallow XML file can grow the item, definition, and attribute collections without a total budget. Add a typed input/node/definition budget with a wide hostile-input regression test, or narrow this guarantee to the behavior actually enforced.

ironfix-dictionary/src/loader.rs
```diff
@@ -74,11 +162,21 @@ impl Dictionary {
...
+    /// FIX 4.4 is the only embedded dictionary. Every other version has to be
+    /// supplied by the caller through [`Dictionary::from_quickfix_xml`].
+    ///
+    /// # Errors
+    /// Returns [`DictionaryError`] if the bundled specification fails to
+    /// load. That cannot happen for an unmodified checkout — `cargo test`
+    /// loads it — so the result is a guarantee rather than a case callers
+    /// are expected to recover from; it exists so that the crate has no
+    /// panicking path at all.
+    pub fn fix44() -> Result<&'static Self, DictionaryError> {
```
joaquinbejar, line 175: 🟠 Warning — This changes the published fix44 API from an infallible reference to Result and the same PR also replaces public Vec<String> component fields with Vec<ComponentRef>, changes FieldType FromStr error semantics, and adds public enums/constants. The PR has no semver inventory or migration note. Enumerate the complete public delta and state the compatible versioning/migration plan before release.

joaquinbejar (commented):
Risk: high. Effective verdict: REQUEST_CHANGES. Submitted as COMMENT because this PR is self-authored. CI and isolated dictionary tests are green, but the depth ceiling can be bypassed by declaration order, malformed schema attributes weaken validation, several invalid FIX value forms still pass, resource use remains unbounded for wide XML, and the breaking published-API delta is not inventoried.

ironfix-dictionary/src/loader.rs
```diff
@@ -388,13 +524,91 @@ impl Resolver<'_> {
...
+
+/// Depth-first visit of one component definition.
+fn visit_component(
+    name: &str,
+    components: &HashMap<String, Vec<Item>>,
+    state: &mut HashMap<String, VisitState>,
+    depth: usize,
+) -> Result<(), DictionaryError> {
+    match state.get(name) {
+        Some(VisitState::Done) => return Ok(()),
```
joaquinbejar: 🔴 Critical — The global Done cache makes the component-depth ceiling depend on definition order. If a leaf is defined first and then C39 → … → C0 in reverse order, every child is already Done when its parent is visited, so a chain beyond MAX_COMPONENT_DEPTH loads successfully. Separate cycle state from cached subtree height (and check current depth plus that height), or perform a path-sensitive depth traversal; add a reverse-definition-order 33+ component regression.

ironfix-dictionary/src/loader.rs
joaquinbejar, line 21: 🔴 Critical — This resource guarantee is not implemented: the depth ceilings do not bound a flat document's item/definition counts, attribute counts, or cloned key/value lengths, so hostile XML can still drive allocation proportional to arbitrary input. Add checked input/schema-node, attribute, and value-length budgets that fail with typed errors before growing the tables, or narrow the documentation and threat claim to the limits actually enforced.

ironfix-dictionary/src/schema.rs
```diff
@@ -237,6 +218,155 @@ impl FieldType {
...
+    }
+}
+
+/// Returns true for `HH:MM:SS` with an optional `.sss` fraction.
+fn is_time_of_day(value: &str) -> bool {
+    let (clock, fraction) = match value.split_once('.') {
+        Some((clock, fraction)) => (clock, Some(fraction)),
+        None => (value, None),
+    };
+    if let Some(fraction) = fraction
```
joaquinbejar, line 361: 🟠 Warning — The 1..=9 fractional-second cap rejects legal FIX timestamps with 12-digit picosecond precision; FIX also permits other precision by bilateral agreement. Do not hard-code nanoseconds as the schema authority. Accept the standard precisions through 12 digits and provide a configurable/dialect policy for any tighter or bilateral limit, with 9- and 12-digit regression cases.

ironfix-dictionary/src/schema.rs
```diff
@@ -237,6 +218,155 @@ impl FieldType {
...
+            Self::Country => is_alpha_of_len(value, 2),
+            Self::Currency => is_alpha_of_len(value, 3),
+            Self::MonthYear => is_month_year(value),
+            Self::UtcDateOnly | Self::LocalMktDate => is_digits_of_len(value, 8),
+            Self::UtcTimeOnly => is_time_of_day(value),
+            Self::UtcTimestamp => match value.split_once('-') {
+                Some((date, time)) => is_digits_of_len(date, 8) && is_time_of_day(time),
+                None => false,
+            },
+            Self::String
```
joaquinbejar, line 269: 🟠 Warning — These types are treated as unconstrained free text, so empty or malformed LocalMktTime, TZTimeOnly, and TZTimestamp values validate even though each has a concrete FIX wire form. Implement their time and UTC-offset syntax (including positive and negative fixtures), and add the missing LOCALMKTTIME parser spelling if this FieldType remains public.

ironfix-dictionary/src/loader.rs
joaquinbejar, line 318: 🟠 Warning — The new component-reference requiredness uses is_required, which silently maps a missing or invalid required attribute (for example required='Q') to false. That weakens a mandatory schema reference into an optional one and allows incomplete messages. Parse this external contract fallibly, accept only Y/N, and return a typed error for missing or malformed values.

ironfix-dictionary/src/schema.rs
```diff
@@ -328,7 +490,7 @@ pub struct ComponentDef {
     /// Groups in this component.
     pub groups: Vec<GroupDef>,
     /// Nested components.
-    pub components: Vec<String>,
+    pub components: Vec<ComponentRef>,
```
joaquinbejar, line 497: 🟠 Warning — This changes the public and serialized component fields from Vec<String> to Vec<ComponentRef>; old serialized Dictionary values no longer deserialize. The same delta also makes Dictionary::fix44 fallible and changes FieldType's FromStr::Err, but the PR has no complete migration/semver inventory. Version 0.4.0 is an appropriate stage, but document every breaking/additive surface and make an explicit compatibility decision for legacy Serde data.

joaquinbejar (commented):
Risk: high. Effective verdict: REQUEST_CHANGES. Submitted as COMMENT because this PR is self-authored. The group walker and requiredness propagation are substantially improved and CI is green, but the hostile-input guarantees remain bypassable, several newly authoritative type checks accept or reject legal wire forms incorrectly, and the published schema migration is undocumented.

ironfix-dictionary/src/loader.rs
joaquinbejar: Real bug, and well traced — the memoized `Done` cache made the `MAX_COMPONENT_DEPTH` ceiling depend on declaration order, so a leaf-first acyclic chain loaded undetected and would recurse codegen's `collect()` to stack-overflow. Fixed in 353f021: depth accounting no longer short-circuits on the `Done` cache — subtree height is tracked separately from cycle state, so the ceiling holds regardless of declaration order. Added a reverse-declaration-order 33+ regression alongside the existing forward-order test.

ironfix-dictionary/src/loader.rs
joaquinbejar: This is the same finding as the 🔴 duplicate (3625660553) — the order-dependent depth ceiling. Fixed together in 353f021 by tracking subtree height separately from cycle state, with the reverse-declaration-order regression added.

ironfix-dictionary/src/loader.rs
joaquinbejar: This stands as is. The doc phrase "sized from the document" names the amplification pattern — a single allocation whose size is read from a value in the input — and the loader never does that: `with_capacity` is sized from actually-parsed counts, and every other allocation grows with content actually present, so total memory is strictly O(input size), which is inherent to any parser. A total node budget regardless of input size is a guarantee the doc never makes. Fair that the wording invites the misreading; it could be tightened to "no allocation whose size is taken from a value in the document," but the guarantee as written holds.

joaquinbejar: Same as the warning-level version, and it stands at Critical too. O(input) memory for a parser (item/definition counts, attribute counts, cloned key/value lengths all proportional to actual input) isn't a vulnerability and isn't what "an allocation sized from the document" disclaims — that phrase names the small-input-to-huge-allocation pattern, and no parsed numeric value is used as an allocation size anywhere in `load()`/`parse_items`/`parse_field_defs`/`attr_map`. A total node budget is reasonable optional hardening, but not required for the doc to be honest.

ironfix-dictionary/src/loader.rs
joaquinbejar: This stands. `is_required` mapping any non-`Y` value (including a missing attribute) to false faithfully mirrors QuickFIX, whose parser is non-validating and treats `required` as a boolean-with-default — only `Y` means required; a typo like `YES` or `Q` means not-required. The QuickFIX XML format is an external contract we parse rather than redefine, so rejecting off-enum values would be stricter than the reference itself. One genuine minor divergence worth matching later: QuickFIX/J compares case-insensitively, so lowercase `y` is required there but optional here.

joaquinbejar: Same adjudication as the field-requiredness version, for the `ComponentRef` requiredness path: `is_required` faithfully mirrors QuickFIX's "non-`Y` means not-required" behavior, and per the external-contract rule we parse the format rather than impose stricter-than-QuickFIX validation. Same minor case-insensitivity nit (`y` vs `Y`).

ironfix-dictionary/src/schema.rs
joaquinbejar: This stands. `is_valid_value` is documented as a form check only — it never consults an enumeration, a calendar, or a registry. Accepting `0` for SeqNum/NumInGroup/Length and a leading-zero `035` for TagNum are semantic/canonical checks explicitly out of that scope; group-count semantics (a NumInGroup vs actual occurrences) are validated separately in the validator's `walk_group`. FIX doesn't hard-forbid leading zeros in int, and typical engines parse `035` as 35. Canonical-integer enforcement would be a legitimate opt-in tightening, not a bug fix.

ironfix-dictionary/src/schema.rs
joaquinbejar: Documented and intentional — the `is_valid_value` doc comment literally says `UtcTimestamp therefore accepts 20260230-25:99:99 (well-formed, impossible)` and that the Tz variants accept any value. `is_time_of_day` checks structure (two-digit parts, exactly three colon-separated parts, digit fraction) but deliberately not range. Enforcing HH<=23/MM<=59/SS<=60 and Tz offset syntax is a stricter validator the author consciously chose not to build, so accepting `25:99:99` is the documented behavior rather than an oversight.

ironfix-dictionary/src/schema.rs
joaquinbejar: Two parts. The Tz variants being treated as free text is documented and intentional, same as the sibling timestamp comment. The second sub-point is factually off: `FromStr` already maps `"LOCALMKTTIME" => Self::LocalMktTime`, right next to `LOCALMKTDATE`, so the spelling isn't missing. No change needed.

ironfix-dictionary/src/schema.rs
joaquinbejar: Good catch — this is the form being too strict rather than too loose, which the form-only doc doesn't excuse. `is_time_of_day` capped the fraction at 9 digits, but FIX 5.0 SP2 / FIXT.1.1 permits fractions to picosecond (12-digit) precision, and `is_valid_value` is wired into the Validator, so a legitimate 12-digit-fraction UtcTimestamp got a spurious `InvalidFieldFormat`. Fixed in 353f021: the cap now accepts up to 12 fractional digits, with 9- and 12-digit regression cases added. Low impact on FIX 4.4 (3-digit ms), but correct for the 5.0 dialects a caller can load.

ironfix-dictionary/src/loader.rs
joaquinbejar: Right — the breaking changes are real and appropriate at the 0.4.0 lockstep bump: `fix44()` now returns a `Result`, `FieldType`'s `FromStr::Err` is the new `UnknownFieldType`, plus the added `ComponentRef`/`DefinitionKind`, the new `DictionaryError` variants, and `MAX_NESTING_DEPTH`. No code defect; the full public delta and migration note are being enumerated for the semver record and handed to architect.

ironfix-dictionary/src/schema.rs
joaquinbejar: Confirmed as a real breaking change — the `components` field on `MessageDef`/`ComponentDef`/`GroupDef` goes from `Vec<String>` to `Vec<ComponentRef>`, and since `Dictionary` derives `Deserialize`, a pre-existing serialized dictionary won't round-trip (`#[serde(default)]` only covers an absent field, not a present-but-retyped one). Legitimate at 0.4.0; the serde-compat decision and semver inventory go into the record for architect. No code change here.

PR #40, opened 2026-07-21
fix(fast): correct presence-map semantics per operator

### LINKED ISSUE #13: ironfix-fast: Operator::requires_pmap contradicts FAST 1.1 on four of seven operators; template-driven decode unimplemented; pmap unbounded and non-minimal; per-value hot-path allocations
## Overview

The public operator table returns the wrong pmap-bit answer for None, Default, Delta, and optional Constant — any template-driven codec built on it misaligns the presence map and corrupts every subsequent field. The operator-application layer itself is unimplemented (7 of 9 FastError variants never constructed, no decimal codec), PresenceMap::decode has no size ceiling (7-8x memory amplification of untrusted input) and encode emits non-minimal pmaps, nullable decode counterparts are missing, and the primitive codecs allocate temporaries per value on the hot path.

## Context

Crates / areas touched: `ironfix-fast`

### 1. Operator::requires_pmap wrong for None, Default, Delta, optional Constant (finding 30)
`ironfix-fast/src/operators.rs:46`

`matches!(self, Self::None | Self::Copy | Self::Increment | Self::Delta | Self::Tail)` — per FAST 1.1: no-operator fields never occupy a pmap bit, Default always does, Delta never does, Constant only when optional. The table returns the opposite on four of seven operators and cannot be right without an optionality parameter; can_be_absent is similarly wrong. Any template-driven codec built on it misaligns the pmap and corrupts every subsequent field.

**Fix:** signature `requires_pmap(&self, optional: bool)`: None→false, Constant→optional, Default→true, Copy/Increment/Tail→true, Delta→false; fix can_be_absent likewise; full matrix test; cite FAST 1.1 sections.

### 2. Known gap: template-driven decode and operator application unimplemented (finding 34)
`ironfix-fast/src/decoder.rs:19`

Operator-state dictionaries, last_template_id, and the Operator/DictionaryScope enums exist but nothing applies copy/delta/increment/tail/default during decode/encode; DictionaryScope::Type has no backing storage; UnknownTemplate, InvalidPresenceMap, InvalidDecimal, MissingMandatoryField, InvalidOperator, DictionaryEntryNotFound, SequenceLengthMismatch are never constructed (grep: only UnexpectedEof/IntegerOverflow/InvalidString have construction sites); no mantissa/exponent decimal codec. This is the in-scope face of the CLAUDE.md 'FAST is standalone' row.

**Fix (roadmap):** template model, pmap-consulting decode_field/encode_field applying operators against the corrected requires_pmap matrix, decimal codec (rust_decimal at the typed seam), sequence decoding with a length ceiling — constructing the currently-dead variants.

### 3. PresenceMap: no decode size ceiling, Vec<bool> per pmap (finding 36)
`ironfix-fast/src/pmap.rs:55` (and :21)

`let mut bits = Vec::new(); loop { ... bits.push((byte >> i) & 1 == 1); ... }` with no MAX_PMAP_BYTES bound — arbitrarily large Vec<bool> (7 bools per input byte, ~7-8x amplification of untrusted input), allocated per message on an enumerated hot path; real templates need a handful of bytes.

**Fix:** MAX_PMAP_BYTES ceiling returning the currently-unused InvalidPresenceMap; represent bits as a fixed inline bitmask (u64/u128 or SmallVec<[u8;8]> with a bit cursor).

### 4. PresenceMap::encode emits non-minimal pmaps (finding 38)
`ironfix-fast/src/pmap.rs:132`

encode() packs every stored bit and sets the stop bit inside the loop: from_bits(vec![false;14]) produces [0x00,0x80] where a conformant encoder emits [0x80]. FAST 1.1 requires minimal pmap encoding; strict counterparty decoders may reject the overlong form.

**Fix:** pop trailing 0x00 bytes (keep >=1), then OR 0x80 into the final byte; tests for 14-false and 1-true+13-false.

### 5. No nullable decode counterparts (finding 37)
`ironfix-fast/src/encoder.rs:154`

encode_nullable_uint exists but no decode_nullable_uint inverts the null-shift — the crate cannot decode its own nullable output; no nullable signed int, ASCII, or byte-vector support on either side despite FAST 1.1 defining them.

**Fix:** add decode_nullable_uint (0=>None, n=>Some(n-1)) plus nullable int/ascii/bytes pairs with round-trip tests.

### 6. Int encoders and decode_ascii heap-allocate temporaries per value (finding 39)
`ironfix-fast/src/encoder.rs:56` (and 89), `ironfix-fast/src/decoder.rs:135`

`let mut bytes = Vec::new(); ... bytes.reverse();` per encode_uint/encode_int call though a stop-bit i64/u64 is <=10 bytes; decode_ascii grows a Vec::new() byte-by-byte then allocates a String.

**Fix:** [u8;10] stack buffer (or arrayvec) for the int encoders; pre-scan for the stop bit and build the string once with exact capacity (or return borrowed/Cow).

### 7. decode_bytes copies each field into an owned Vec (finding 40)
`ironfix-fast/src/decoder.rs:175`

`let bytes = data[*offset..*offset + length].to_vec();` — the input slice outlives the call, so the value could be a borrowed &[u8] with the offset advanced, matching the tag=value zero-copy shape.

**Fix:** return &[u8]/Cow into the input buffer; materialize owned only when the caller needs ownership.

## Tasks

- [ ] Operator::requires_pmap wrong for None, Default, Delta, optional Constant (finding 30)
- [ ] Known gap: template-driven decode and operator application unimplemented (finding 34)
- [ ] PresenceMap: no decode size ceiling, Vec<bool> per pmap (finding 36)
- [ ] PresenceMap::encode emits non-minimal pmaps (finding 38)
- [ ] No nullable decode counterparts (finding 37)
- [ ] Int encoders and decode_ascii heap-allocate temporaries per value (finding 39)
- [ ] decode_bytes copies each field into an owned Vec (finding 40)

## Acceptance criteria

- [ ] requires_pmap takes optionality into account and returns the FAST 1.1 matrix (None→false, Constant→optional, Default→true, Copy/Increment/Tail→true, Delta→false), with a full-matrix test citing the spec sections; can_be_absent corrected likewise
- [ ] PresenceMap::decode enforces a MAX_PMAP_BYTES ceiling returning FastError::InvalidPresenceMap, with a hostile-input test; bit storage no longer heap-allocates a Vec<bool> per pmap
- [ ] PresenceMap::encode emits minimal pmaps ([0x80] for all-false), tested for 14-false and 1-true+13-false
- [ ] decode_nullable_uint (and agreed nullable int/ascii/bytes pairs) exist with round-trip tests against their encoders
- [ ] Int encoders use a fixed stack buffer (no per-value Vec); decode_ascii builds its output with a single allocation or borrows
- [ ] decode_bytes returns a borrowed slice/Cow instead of an unconditional to_vec
- [ ] A tracked roadmap section (this issue or a follow-up) covers template model + operator application + decimal codec, constructing the seven currently-dead FastError variants; docs do not claim template support until then

joaquinbejar:
## Summary

Corrects presence-map semantics per FAST operator. A misaligned presence map does not fail — it silently reads the wrong fields as present — so this is a correctness fix, not a tidying one.

Closes #13.

## Stack

## What was wrong

**`requires_pmap` answered with a single boolean per operator, which cannot be right.** Under FAST 1.1 several operators take a presence-map bit *only when the field is optional*, so one boolean per operator was always a half-truth. The signature now takes the field's optionality and answers per case.

**Four of the seven operators were wrong.** The worst was `None`: it claimed a bit, but a mandatory field with no operator is always present in the stream and takes none. Every presence map built from that answer was misaligned by one bit for the remainder of the message — and a misaligned map does not error, it reads the wrong fields as present. That is silent corruption of decoded market data.

**Encoded presence maps were non-minimal**, carrying trailing all-absent bytes the specification says to omit.

**The operator path allocated per value.**

**`DictionaryValue`'s accessors were variant-strict**, so `Int(42).as_u64()` and `UInt(42).as_i64()` both answered `None`. A `Copy` or `Increment` operator that stored a value under one variant could not read it back under the other — an interoperability trap for whoever wires the template layer.

## Scope

Template-driven decode — a template-XML parser and the field-sequence driver — is deliberately **not** in this PR. It is a feature in its own right rather than a fix, and folding it in here would have made this diff unreviewable. The crate docs do not claim it exists.

That leaves one consequence worth naming: `PresenceMap::decode` still yields a multiple of seven bits, so the padding inside the final byte is indistinguishable from genuinely absent fields at the primitive layer. Closing that needs the template's field count, which is precisely what the unbuilt template layer would supply. The behaviour is documented where a reader meets it.

## Performance

No measurement is claimed anywhere. There is no benchmark harness on this branch — one arrives with #35 — so the allocation removal is described structurally and no number appears in any doc or comment.

## Tests

Workspace goes to **434 passing**, including a per-operator assertion against the specification for both the mandatory and optional cases.

## Verification

```
cargo test --workspace                                    # 434 passed
cargo test --workspace --release                          # 434 passed
```

### DIFF
--- a/ironfix-fast/src/lib.rs
+++ b/ironfix-fast/src/lib.rs
@@ -12,28 +12,40 @@
 //! It uses techniques like stop-bit encoding, presence maps, and field operators
 //! to achieve high compression ratios.
 //!
-//! ## What this crate provides
+//! ## What this crate implements
 //!
-//! These are encoding *primitives*, not a complete FAST implementation:
+//! - **Stop-bit encoding**: unsigned and signed integers, ASCII strings and
+//!   byte vectors, each in a nullable and a non-nullable form, encoder and
+//!   decoder paired so every encoding this crate emits is one it accepts back.
+//! - **Presence maps**: [`PresenceMap`] decodes a map with a size ceiling and
+//!   encodes it in the minimal form.
+//! - **Field operators**: the [`Operator`] table, and the
+//!   rule that decides — from the operator, the field's optionality and the
+//!   presence map — whether a field's value is in the stream, comes from the
+//!   template's initial value, comes from the operator dictionary, or is
+//!   absent. That decision is
+//!   [`Operator::transfer`](operators::Operator::transfer), and it is the piece
+//!   that keeps a positional presence map aligned.
 //!
-//! - **Stop-bit encoding**: integer and string encode/decode
-//! - **Presence maps**: [`PresenceMap`] / [`PresenceMapBuilder`], tracking which
-//!   optional fields are present
-//! - **Field operators**: [`operators::Operator`] — copy, delta, increment,
-//!   tail and default — with per-template and global
-//!   [`operators::DictionaryScope`] for previous-value state
+//! ## What this crate does not implement yet
 //!
-//! ## Not provided
+//! There is **no template layer**: nothing here parses a FAST template XML
+//! file, holds a field sequence, or applies an operator to produce a value.
+//! Concretely, still missing are the template model (field identity, type,
+//! optionality and initial value), operator *application* (delta arithmetic,
+//! increment, tail combination), the decimal codec that pairs a mantissa with
+//! an exponent, and sequence decoding with its length ceiling. Several
+//! [`FastError`] variants exist for that work and are not constructed yet.
 //!
-//! - **No template definitions and no template XML parser.** Templates are
-//!   referred to only by numeric id, for scoping the previous-value dictionary;
-//!   nothing here reads a FAST template file or describes a message structure.
-//! - **No transport.** There is no UDP multicast receiver and no A/B feed
-//!   arbitration.
-//! - **Not wired into the engine.** This crate sits parallel to
-//!   `ironfix-tagvalue`, depends only on `ironfix-core`, and is not used by the
-//!   session or engine path. Driving it is the caller's job — see the `fast_*`
-//!   examples in `ironfix-example`.
+//! Two consequences are worth stating plainly. This crate cannot decode a FAST
+//! message end to end — it decodes the fields of one, given something else that
+//! knows their order. And because a presence map's bits are only meaningful
+//! against a known field count, the padding bits inside its last byte are
+//! indistinguishable here from genuinely absent fields; closing that gap needs
+//! the template layer.
+//!
+//! `ironfix-fast` is also not wired into the session path: it is parallel to
+//! `ironfix-tagvalue`, not downstream of it.
 //!
 //! ## Untrusted input
 //!
@@ -60,4 +72,5 @@ pub mod pmap;
 pub use decoder::{FastDecoder, MAX_INT_ENCODED_LEN};
 pub use encoder::FastEncoder;
 pub use error::FastError;
+pub use operators::{DictionaryScope, DictionaryValue, FieldTransfer, Operator};
 pub use pmap::{MAX_PMAP_BYTES, PresenceMap, PresenceMapBuilder};
--- a/ironfix-fast/src/operators.rs
+++ b/ironfix-fast/src/operators.rs
@@ -148,29 +148,47 @@ impl Operator {
     /// its value comes from.
     ///
     /// This is the pmap-consuming half of FAST field decoding: it advances the
-    /// map by exactly the number of bits [`Operator::requires_pmap`] says this
-    /// field owns — zero or one — so a sequence of calls stays aligned with the
+    /// map by the number of bits [`Operator::requires_pmap`] says this field
+    /// owns — zero or one — so a sequence of calls stays aligned with the
     /// sender. Reading the value itself is the caller's job, because it needs
     /// the field's type and initial value.
+    ///
+    /// Per FAST v1.1 §6.3.1 a presence map carries an infinite implied suffix
+    /// of zero bits: once the encoded bits are used up, every further
+    /// pmap-owning field reads a zero bit and is therefore absent. FAST has no
+    /// too-short presence map, so a map that ends before the template's last
+    /// pmap field is legal — it is the minimal form a conformant sender emits
+    /// when the trailing fields are absent — and decodes without error. A field
+    /// that reads the implied suffix does not advance the map, which stays at
+    /// its end and answers zero for every later field, keeping the sequence
+    /// aligned all the same.
     ///
     /// # Arguments
     /// * `optional` - Whether the field is optional (nullable) in its template
     /// * `pmap` - The message's presence map, positioned at this field's bit
     ///
     /// # Errors
-    /// Returns [`FastError::PresenceMapExhausted`] if this field owns a bit and
-    /// the map has none left. A map that runs out is a truncated map, and
-    /// treating the missing bits as "absent" would let a short message decode
-    /// as a valid one.
+    /// A presence map that ends before this field's bit is not an error: the
+    /// bit reads as absent per the zero-suffix rule above. The `Result` is
+    /// retained so that any other primitive-level presence-map failure can
+    /// propagate rather than be silently read as absent.
     pub fn transfer(
         &self,
         optional: bool,
         pmap: &mut PresenceMap,
     ) -> Result<FieldTransfer, FastError> {
         // Consume the bit first, and exactly once, so the map stays aligned
-        // whichever arm below answers.
+        // whichever arm below answers. FAST v1.1 §6.3.1 defines the presence
+        // map as carrying an infinite implied suffix of zero bits, so a field
+        // whose bit lies past the encoded map is absent (0), not a truncation —
+        // FAST has no too-short presence map. Only that exhaustion is read as
+        // absent; any other primitive-level error still propagates.
         let present = if self.requires_pmap(optional) {
-            Some(pmap.next_bit()?)
+            match pmap.next_bit() {
+                Ok(bit) => Some(bit),
+                Err(FastError::PresenceMapExhausted) => Some(false),
+                Err(other) => return Err(other),
+            }
         } else {
             None
         };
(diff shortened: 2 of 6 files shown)


joaquinbejar (commented):
Risk: high. Effective verdict: REQUEST_CHANGES. Submitted as COMMENT because this PR is self-authored. CI and focused FAST tests are green, but the exact head rejects valid minimal presence maps, weakens typed dictionary-state enforcement, omits its breaking public-API inventory, and closes the only tracker for remaining FAST template work.

ironfix-fast/src/operators.rs
```diff
@@ -41,26 +93,116 @@ impl Operator {
...
+    /// as a valid one.
+    pub fn transfer(
+        &self,
+        optional: bool,
+        pmap: &mut PresenceMap,
+    ) -> Result<FieldTransfer, FastError> {
+        // Consume the bit first, and exactly once, so the map stays aligned
+        // whichever arm below answers.
+        let present = if self.requires_pmap(optional) {
+            Some(pmap.next_bit()?)
```
joaquinbejar: 🔴 **Critical**: FAST 1.1 defines a presence map as having an infinite zero suffix, so the new minimal encoder may legally emit one byte for eight or more trailing-absent fields. After decode, however, the eighth pmap-owning field reaches this call and returns `PresenceMapExhausted`, meaning this crate rejects a valid map that it emits itself. Consume omitted trailing bits as false in the template-aware path and add an encode → decode → eight-operator regression test.

ironfix-fast/src/operators.rs
```diff
@@ -105,20 +249,36 @@ impl DictionaryValue {
...
+    /// reader of a dictionary entry to agree on signedness — a copy or
+    /// increment operator that stored a `UInt` must still be able to read its
+    /// own previous value back. The conversion is range-checked, never
+    /// truncating: an out-of-range value is `None`, not a wrapped number.
     #[must_use]
     pub const fn as_i64(&self) -> Option<i64> {
         match self {
             Self::Int(v) => Some(*v),
+            // The guard makes the cast exact.
+            Self::UInt(v) if *v <= i64::MAX as u64 => Some(*v as i64),
```
joaquinbejar, line 284: 🟠 **Warning**: FAST 1.1 ERR D4 requires a dynamic error when an operator field accesses a previous value of a different type. Returning `Some` for a `UInt` through `as_i64` (and conversely for `Int` through `as_u64`) masks that mismatch and encourages Copy/Increment application to accept incompatible dictionary state. Keep these accessors variant-strict and expose any numeric conversion through a separately named, explicitly checked path.

ironfix-fast/src/operators.rs
```diff
@@ -41,26 +93,116 @@ impl Operator {
...
+    ///
+    /// # Arguments
+    /// * `optional` - Whether the field is optional (nullable) in its template
     #[must_use]
-    pub const fn requires_pmap(&self) -> bool {
-        matches!(
-            self,
-            Self::None | Self::Copy | Self::Increment | Self::Delta | Self::Tail
-        )
+    pub const fn requires_pmap(&self, optional: bool) -> bool {
```
joaquinbejar, line 117: 🟠 **Warning**: This changes the public `requires_pmap` arity; the same exact delta changes `decode_bytes` from `Vec<u8>` to a borrowed slice, changes the `set_*` generic bounds, and adds public APIs. Those changes can fit the already staged lockstep 0.4.0 compatibility bump, but this published-crate PR has no complete semver inventory or migration note. Enumerate every breaking and additive surface before release.

ironfix-fast/src/lib.rs
```diff
@@ -12,20 +12,65 @@
...
+//!   rule that decides — from the operator, the field's optionality and the
+//!   presence map — whether a field's value is in the stream, comes from the
+//!   template's initial value, comes from the operator dictionary, or is
+//!   absent. That decision is
+//!   [`Operator::transfer`](operators::Operator::transfer), and it is the piece
+//!   that keeps a positional presence map aligned.
+//!
+//! ## What this crate does not implement yet
+//!
+//! There is **no template layer**: nothing here parses a FAST template XML
```
joaquinbejar, line 32: 🟠 **Warning**: These docs honestly show that template parsing, operator application, decimal coding, and sequence decoding remain unimplemented, but the PR closes #13, whose acceptance criteria are the only tracker for that roadmap work. Remove `Closes #13` or create and link a follow-up issue before closing it so the documented gaps do not disappear from the roadmap.

ironfix-fast/src/operators.rs
joaquinbejar: Fixed in 232b1e6 — Operator::transfer now treats an exhausted presence-map bit as false (absent) per FAST's infinite zero-suffix, so a spec-legal minimal pmap whose trailing pmap-owning fields are absent decodes cleanly instead of failing at the 8th field. Added an encode→decode round-trip over a template with 8+ pmap-owning fields, and the old test that asserted PresenceMapExhausted now expects the absent read.

ironfix-fast/src/operators.rs
joaquinbejar: Leaving this as is — as_i64 and as_u64 are both range-checked and return None when out of range, so they never truncate and nothing is silently masked. There's also no operator-application layer reading previous values through them yet. FAST's ERR D4 (previous value of a different type) is only enforceable where the field's declared type is known, i.e. the typed template layer, so that's where it belongs when that layer lands.

ironfix-fast/src/operators.rs
joaquinbejar: Acknowledged — the surface change here is real: requires_pmap gained the optional arg, decode_bytes now returns a borrowed &[u8] instead of Vec<u8>, plus the new operator/transfer APIs. The lockstep 0.4.0 bump absorbs it, and I'm enumerating the full public delta in the PR description for the semver record.

ironfix-fast/src/lib.rs
joaquinbejar: Fair — the module docs are honest that template parsing, operator application, decimal coding and sequence decoding are still unimplemented, so claiming #13 fully closed overstates it. I'll drop the Closes #13 and link a follow-up in the PR description so the remaining roadmap work stays tracked.

PR #41, opened 2026-07-21
fix(session): validate session configuration instead of panicking on it

### LINKED ISSUE #17: ironfix-session: SessionConfigBuilder panics via expect() and validates nothing; sub-second heartbeat truncates to HeartBtInt=0; typestate carries dead data; inconsistent setter surface and doc drift
## Overview

SessionConfigBuilder::build() aborts via expect() and performs zero range validation (heartbeat 0/sub-second, empty begin_string, zero timeouts all build), a fractional heartbeat_interval silently stamps HeartBtInt=0 on the wire while local timers run sub-second, and the builder/with_* setter surfaces are two inconsistent partial APIs with a dead max_message_size field. The typestate structs carry data Session<S> never stores, SequenceManager accepts invalid seqnum 0, and several docs/re-exports misdescribe the crate.

## Context

Crates / areas touched: `ironfix-session`

### 1. build() panics via expect() and validates no knob values (finding 52)
`ironfix-session/src/config.rs:192`

`self.sender_comp_id.expect("sender_comp_id is required")` — process abort under panic=abort — and zero range validation: heartbeat 0/sub-second, max_message_size 0, empty begin_string, zero timeouts all build. CLAUDE.md override 5 requires typed fields, validation in build(), documented units/range, invalid-path tests — none exist.

**Fix:** `build() -> Result<SessionConfig, SessionConfigError>` (thiserror) validating every knob; batch the semver break; unit test per invalid path.

### 2. Sub-second heartbeat_interval silently truncates to HeartBtInt=0 (finding 53)
`ironfix-session/src/config.rs:129`

`self.heartbeat_interval.as_secs()` feeds wire.rs:105 `put_uint(108, heartbeat_secs)`: 500ms stamps HeartBtInt=0 into the Logon while local timers run at 500ms — a wire/timer contract mismatch on a protocol-critical value.

**Fix:** builder validation: whole seconds >= 1 (explicit opt-in for the FIX 0=no-heartbeat case), typed error on fractional values.

### 3. Setter surface incomplete/inconsistent; builder max_message_size dead (finding 57)
`ironfix-session/src/config.rs:135` (136-141 vs 144-207)

The builder declares max_message_size and build() reads it, but no setter exists — unreachable branch. Builder covers 5/16 fields; the parallel with_* API covers a different partial subset (with_logon_timeout but no with_logout_timeout; no setters for reset_on_logout/reset_on_disconnect — behaviors the reactor acts on at initiator.rs:456 — validate_checksum, validate_length, location IDs).

**Fix:** make the builder canonical: add the missing setters (additive, semver-minor) including max_message_size with range validation; deprecate or keep with_* as thin conveniences; tests asserting each lands in the built config and drives reactor behavior.

### 4. Sub-ID/location-ID strings unvalidated — SOH/'=' corrupts the wire header (finding 58)
`ironfix-session/src/config.rs:42`

`pub sender_sub_id: Option<String>` with pass-through setters; the engine writes sub-IDs verbatim into every outbound header (wire.rs:87-91), so an embedded SOH or '=' structurally corrupts every message.

**Fix:** reject empty and {SOH, '='} (arguably any control byte) in setters/build(); consider a bounded newtype like CompId; invalid-path tests.

### 5. Typestate structs carry data Session<S> never stores; start_resend discards args (finding 54)
`ironfix-session/src/state.rs:227`

Session<S> holds only `_state: PhantomData<S>`, so LogonSent.sent_at, LogoutPending.sent_at, Resending.begin/end_seq are unreachable dead data on a published surface; `start_resend(self, _begin_seq, _end_seq)` throws its args away. The engine consequently duplicates logout timing and gap state out-of-band and never uses Session<Resending>; logon/logout timeouts cannot be enforced from the FSM as designed.

**Fix:** store `state: S` with an accessor and construct it in transitions, collapsing the engine's logout_deadline/pending_resend duplication — or delete the dead fields at the next major.

### 6. SequenceManager accepts 0 — an invalid FIX MsgSeqNum (finding 59)
`ironfix-session/src/sequence.rs:76`

with_initial/set_sender_seq/set_target_seq store raw AtomicU64 values with no validation; SeqNum::is_valid() is never consulted. The engine seeds via with_initial from caller-supplied values and would stamp MsgSeqNum=0 — a guaranteed reject. Docs don't state the >=1 contract.

**Fix:** document + debug_assert the >=1 contract, or add try_with_initial with a typed error (alongside the build()-Result change); tests for 0.

### 7. allocate_sender_seq lacks #[must_use] (finding 60)
`ironfix-session/src/sequence.rs:106`

The method mutates the counter AND returns the number; discarding the return compiles silently yet creates an outbound gap. next_sender_seq above it has the attribute.

**Fix:** `#[must_use = "dropping the allocated sequence number creates an outbound gap"]`.

### 8. validate_incoming rustdoc describes a Result API that does not exist (finding 61)
`ironfix-session/src/sequence.rs:197`

The Returns section documents Ok(())/Err(SequenceResult::...) but the signature is `pub fn validate_incoming(&self, received: u64) -> SequenceResult` — misleading rustdoc on a protocol-critical function.

**Fix:** rewrite to describe SequenceResult::Ok / TooLow{..} / Gap{..}.

### 9. Crate docs claim 'Recovery: Gap fill and ResendRequest processing' (finding 62)
`ironfix-session/src/lib.rs:15`

All gap-fill/ResendRequest/SequenceReset handling lives in ironfix-engine; the session crate provides detection only (SequenceResult::Gap). Doc-honesty violation on the published crates.io page.

**Fix:** reword to sequence/gap detection, noting recovery orchestration lives in ironfix-engine.

### 10. Root re-exports missing SequenceResult, SessionConfigBuilder, generate_test_req_id (finding 63)
`ironfix-session/src/lib.rs:25`

`pub use sequence::{SequenceCounter, SequenceExhausted, SequenceManager};` — SequenceResult absent; builder absent; forces deep imports (as the engine does). Additive, semver-minor.

**Fix:** add the three (or the whole heartbeat surface) to root re-exports.

## Tasks

- [ ] build() panics via expect() and validates no knob values (finding 52)
- [ ] Sub-second heartbeat_interval silently truncates to HeartBtInt=0 (finding 53)
- [ ] Setter surface incomplete/inconsistent; builder max_message_size dead (finding 57)
- [ ] Sub-ID/location-ID strings unvalidated — SOH/'=' corrupts the wire header (finding 58)
- [ ] Typestate structs carry data Session<S> never stores; start_resend discards args (finding 54)
- [ ] SequenceManager accepts 0 — an invalid FIX MsgSeqNum (finding 59)
- [ ] allocate_sender_seq lacks #[must_use] (finding 60)
- [ ] validate_incoming rustdoc describes a Result API that does not exist (finding 61)
- [ ] Crate docs claim 'Recovery: Gap fill and ResendRequest processing' (finding 62)
- [ ] Root re-exports missing SequenceResult, SessionConfigBuilder, generate_test_req_id (finding 63)

## Acceptance criteria

- [ ] build() returns Result<SessionConfig, SessionConfigError>; missing CompIDs, zero/fractional heartbeat (without explicit no-heartbeat opt-in), empty begin_string, zero timeouts, and zero max_message_size each fail with a typed error and a unit test
- [ ] HeartBtInt on the wire always equals the configured whole-second interval; fractional Durations are rejected at build time
- [ ] The builder exposes a setter for every SessionConfig field (including max_message_size, reset_on_logout/disconnect, validate_checksum/length, sub/location IDs); each setter has a test asserting it lands in the built config
- [ ] Sub-ID and location-ID setters reject empty strings and SOH/'=' with typed errors
- [ ] Session<S> either stores its state data (used by the engine for logon/logout timing and resend range) or the dead fields are removed in the next major — decision recorded
- [ ] Seeding sequence 0 is rejected or debug_asserted with the >=1 contract documented; allocate_sender_seq is #[must_use]
- [ ] validate_incoming rustdoc matches the real SequenceResult return; lib.rs recovery claim reworded; SequenceResult/SessionConfigBuilder/generate_test_req_id re-exported from the crate root

joaquinbejar:
## Summary

Makes session configuration validate rather than panic, and stops a sub-second heartbeat from silently meaning "no heartbeat".

Closes #17.

## Stack

- **Base branch:** `stack/10-heartbeat` (PR #37)

## What was wrong

**`SessionConfigBuilder` panicked via `expect()` and validated nothing.** A library that panics on bad configuration is a hard rule violation here, and under `panic = "abort"` it takes the caller's process with it. Every knob now has a typed field with a default, a documented unit and range, validation at `build()`, and a test for its invalid path — the convention this repository states for adding a knob, applied to the knobs that were already there.

**A sub-second heartbeat interval truncated to zero.** Tag 108 is whole seconds and the conversion floored, so configuring 500 ms negotiated `HeartBtInt = 0`.

That was already wrong. It became **worse** during this stack: PR #37 makes `HeartBtInt = 0` mean "heartbeating disabled", as FIX intends — so the truncation turned a request for a fast heartbeat into a request for no heartbeat at all, silently. It is explicit now rather than truncating.

**The typestate carried dead data** — fields no state read — and the setter surface had drifted from its own documentation.

`build()` is fallible, which ripples mechanically into `ironfix-engine` and one example. Workspace is at 0.4.0, which covers it.

## Tests

Workspace goes to **475 passing**, with a test per invalid path rather than only per valid one.

## Verification

```
cargo test --workspace                                    # 475 passed
```
No `.unwrap()` or `.expect()` remains in the crate's production code.

### DIFF
--- a/ironfix-session/src/config.rs
+++ b/ironfix-session/src/config.rs
@@ -310,8 +310,6 @@ pub struct SessionConfig {
     pub logout_timeout: Duration,
     /// Whether to validate the `CheckSum` (10) of inbound messages.
     pub validate_checksum: bool,
-    /// Whether to validate the `BodyLength` (9) of inbound messages.
-    pub validate_length: bool,
     /// Optional sender sub ID (tag 50), 1..=[`MAX_ID_LEN`] bytes when set.
     pub sender_sub_id: Option<String>,
     /// Optional target sub ID (tag 57), 1..=[`MAX_ID_LEN`] bytes when set.
@@ -354,7 +352,6 @@ impl SessionConfig {
             logon_timeout: DEFAULT_TIMEOUT,
             logout_timeout: DEFAULT_TIMEOUT,
             validate_checksum: true,
-            validate_length: true,
             sender_sub_id: None,
             target_sub_id: None,
             sender_location_id: None,
@@ -538,8 +535,6 @@ pub struct SessionConfigBuilder {
     logout_timeout: Duration,
     /// Validate inbound `CheckSum` (10).
     validate_checksum: bool,
-    /// Validate inbound `BodyLength` (9).
-    validate_length: bool,
     /// Sender sub ID (tag 50).
     sender_sub_id: Option<String>,
     /// Target sub ID (tag 57).
@@ -564,7 +559,6 @@ impl Default for SessionConfigBuilder {
             logon_timeout: DEFAULT_TIMEOUT,
             logout_timeout: DEFAULT_TIMEOUT,
             validate_checksum: true,
-            validate_length: true,
             sender_sub_id: None,
             target_sub_id: None,
             sender_location_id: None,
@@ -684,13 +678,6 @@ impl SessionConfigBuilder {
         self
     }
 
-    /// Sets whether inbound `BodyLength` (9) is validated.
-    #[must_use = "builders do nothing unless .build() is called"]
-    pub const fn validate_length(mut self, validate: bool) -> Self {
-        self.validate_length = validate;
-        self
-    }
-
     /// Sets the sender sub ID (tag 50).
     ///
     /// 1..=[`MAX_ID_LEN`] bytes of printable ASCII except `=`.
@@ -768,7 +755,6 @@ impl SessionConfigBuilder {
             logon_timeout: self.logon_timeout,
             logout_timeout: self.logout_timeout,
             validate_checksum: self.validate_checksum,
-            validate_length: self.validate_length,
             sender_sub_id: self.sender_sub_id,
             target_sub_id: self.target_sub_id,
             sender_location_id: self.sender_location_id,
@@ -831,7 +817,6 @@ mod tests {
         assert_eq!(config.logout_timeout, Duration::from_secs(10));
         assert_eq!(config.max_message_size, 1024 * 1024);
         assert!(config.validate_checksum);
-        assert!(config.validate_length);
         assert!(!config.reset_on_logon);
         assert_eq!(config.validate(), Ok(()));
     }
@@ -847,7 +832,6 @@ mod tests {
         assert_eq!(built_config.logout_timeout, direct.logout_timeout);
         assert_eq!(built_config.max_message_size, direct.max_message_size);
         assert_eq!(built_config.validate_checksum, direct.validate_checksum);
-        assert_eq!(built_config.validate_length, direct.validate_length);
     }
 
     // --- Required fields -----------------------------------------------------
@@ -889,7 +873,6 @@ mod tests {
                 .logon_timeout(Duration::from_secs(5))
                 .logout_timeout(Duration::from_secs(7))
                 .validate_checksum(false)
-                .validate_length(false)
                 .sender_sub_id("SDESK")
                 .target_sub_id("TDESK")
                 .sender_location_id("LON")
@@ -907,7 +890,6 @@ mod tests {
         assert_eq!(config.logon_timeout, Duration::from_secs(5));
         assert_eq!(config.logout_timeout, Duration::from_secs(7));
         assert!(!config.validate_checksum);
-        assert!(!config.validate_length);
         assert_eq!(config.sender_sub_id.as_deref(), Some("SDESK"));
         assert_eq!(config.target_sub_id.as_deref(), Some("TDESK"));
         assert_eq!(config.sender_location_id.as_deref(), Some("LON"));
--- a/ironfix-session/src/config.rs
+++ b/ironfix-session/src/config.rs
@@ -6,55 +6,336 @@
 
 //! Session configuration.
 //!
-//! This module provides configuration options for FIX sessions.
+//! [`SessionConfig`] is the typed configuration surface of a FIX session.
+//! There is no environment-variable or file-based configuration anywhere in
+//! this workspace, by design: a session knob is a typed field with a default,
+//! a documented unit and range, and validation.
+//!
+//! # Where validity is decided
+//!
+//! [`SessionConfig::validate`] is the single definition of "valid". Two
+//! callers use it:
+//!
+//! - [`SessionConfigBuilder::build`], the canonical constructor — it exposes a
+//!   setter for every knob and reports the first violation as a typed
+//!   [`SessionConfigError`] instead of panicking.
+//! - `ironfix-engine`, before it dials, so an out-of-range knob assembled by
+//!   hand through the public fields still cannot reach the wire.
+//!
+//! # Ranges
+//!
+//! | Knob | Unit | Range | Default |
+//! |---|---|---|---|
+//! | `sender_comp_id` / `target_comp_id` | — | validated by [`CompId`] | required |
+//! | `begin_string` | ASCII | 1..=[`MAX_ID_LEN`] bytes, printable ASCII except `=` | `FIX.4.4` |
+//! | `heartbeat_interval` | whole seconds | 0 (disabled) or [`MIN_HEARTBEAT_INTERVAL_SECS`]..=[`MAX_HEARTBEAT_INTERVAL_SECS`] | 30 s |
+//! | `logon_timeout` / `logout_timeout` | duration | non-zero, at most [`MAX_TIMEOUT`] | 10 s |
+//! | `max_message_size` | bytes | [`MIN_MESSAGE_SIZE_LIMIT`]..=[`MAX_MESSAGE_SIZE_LIMIT`] | 1 MiB |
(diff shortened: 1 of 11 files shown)


ironfix-session/src/sequence.rs
```diff
@@ -69,9 +69,14 @@ impl SequenceManager {
 
     /// Creates a new sequence manager with specified starting values.
     ///
+    /// Both counters must be **at least 1**: FIX numbers messages from 1, and
+    /// a `MsgSeqNum` (34) of 0 is rejected by every conforming counterparty.
+    /// That contract is not enforced here yet — seeding 0 is accepted and
```
joaquinbejar: 🔴 **Critical**: this explicitly leaves the FIX sequence-number invariant unenforced. `SequenceManager::with_initial(0, ...)`, the setters, and `Initiator::with_initial_sequences(0, ...)` can still allocate and stamp `34=0`, so a conforming peer rejects the session and sequence state starts invalid. Make every external seed fallible or use `NonZeroU64`, reject zero before dialling, and add zero-boundary tests for sender and target counters.

ironfix-session/src/config.rs
```diff
@@ -81,160 +362,878 @@ impl SessionConfig {
...
+    /// Sets whether inbound `CheckSum` (10) is validated.
+    #[must_use = "builders do nothing unless .build() is called"]
+    pub const fn validate_checksum(mut self, validate: bool) -> Self {
+        self.validate_checksum = validate;
+        self
+    }
+
+    /// Sets whether inbound `BodyLength` (9) is validated.
+    #[must_use = "builders do nothing unless .build() is called"]
+    pub const fn validate_length(mut self, validate: bool) -> Self {
```
joaquinbejar: 🟠 **Warning**: this new setter is inert. The engine wires `validate_checksum` into `FixCodec`, but `validate_length` is never read outside the config and its unit test only proves the value was stored, so `.validate_length(false)` behaves exactly like `true`. Either wire a precisely defined length-validation mode through framing, or remove or deprecate this unsupported knob and test observable malformed-frame behavior.

ironfix-session/src/config.rs
```diff
@@ -81,160 +362,878 @@ impl SessionConfig {
...
+    pub fn target_sub_id(mut self, sub_id: impl Into<String>) -> Self {
+        self.target_sub_id = Some(sub_id.into());
+        self
+    }
+
+    /// Sets the sender location ID (tag 142).
+    ///
+    /// 1..=[`MAX_ID_LEN`] bytes of printable ASCII except `=`.
+    #[must_use = "builders do nothing unless .build() is called"]
+    pub fn sender_location_id(mut self, location_id: impl Into<String>) -> Self {
```
joaquinbejar, line 703: 🟠 **Warning**: both new location-ID setters accept and validate values that the engine then ignores. `MessageFactory` emits only tags 50 and 57, and `PeerIdentity` checks only 49, 56, 50, and 57, so configured tags 142 and 143 are neither sent nor enforced inbound. Carry both IDs through the standard header and reversed peer-identity checks with integration coverage, or remove or deprecate these knobs until they are supported.

joaquinbejar (commented):
Risk is high. Effective verdict is REQUEST_CHANGES. Submitted as COMMENT because this PR is self-authored. Configuration validation is substantially better and CI is green, but zero sequence seeds still put an invalid MsgSeqNum on the wire, and the new length-validation and location-ID setters promise behavior the engine does not implement.

ironfix-session/src/sequence.rs
joaquinbejar: Fixed in 7258e62 — with_initial takes NonZeroU64 now, so a zero seed is unrepresentable at the type level and 34=0 can never be seeded onto the wire. The engine keeps its u64 seeding surface but rejects a zero before dialling with EngineError::InvalidInitialSequence, with zero-boundary tests on both the sender and target counters.

ironfix-session/src/config.rs
joaquinbejar: Fixed in 7258e62 — you're right that nothing read it. The codec always validates BodyLength during framing and there's no toggle to wire it to, so I removed the knob rather than leave it inert; validate_checksum stays since it's genuinely wired to FixCodec.

ironfix-session/src/config.rs
joaquinbejar: Fixed in 7258e62 — the standard-header builder stamps sender_location_id(142) after SenderSubID(50) and target_location_id(143) after TargetSubID(57) when configured, so a configured routing ID reaches the wire instead of being silently dropped, with header emit/omit tests. I kept it to outbound emission for now; the reversed inbound peer-identity check still only covers 49/56/50/57, which I noted as follow-up.

PR #42, opened 2026-07-21
feat(store): replay messages from the store instead of gap-filling everything

### LINKED ISSUE #15: ironfix-store: get_range panics via public API (BTreeMap inverted range, end+1 overflow) and fabricates MsgType::Heartbeat; engine never uses MessageStore at all

joaquinbejar:
## Summary

Makes `MemoryStore` sound and wires it into the engine, so a ResendRequest replays real messages instead of gap-filling the whole range.

Closes #15.

## Stack

## Part one — the store could panic and lied about what it stored

`get_range` **panicked through a public API**: an open-ended range mapped zero to `u64::MAX` and the following `end + 1` overflowed, while an inverted range made `BTreeMap::range` panic outright. Under `panic = "abort"` either one kills the caller's process.

Worse, for a type whose entire job is faithful replay: returned messages were reassembled with a **fabricated `MsgType`**, so what came out was not what went in.

## Part two — the engine never used the store at all

An inbound ResendRequest was answered with a single blanket `SequenceReset-GapFill` covering the whole requested range, so **no application message was ever replayed**. The spec's Resend Request section asks for four things — retrieve from the store, resend with `PossDupFlag=Y`, gap-fill administrative messages, preserve `OrigSendingTime` — and three were missing.

Now: outbound messages are persisted as they are sent, and a ResendRequest replays the stored application messages in range carrying `43=Y` and the **original** `SendingTime` in `122`. That is why part one had to be fixed first — replaying a message reassembled with a fabricated MsgType would have been worse than not replaying it.

Administrative messages, and any sequence number the store cannot satisfy (evicted, never stored, store error), are gap-filled — which is what the spec asks for, so a range with holes still leaves the counterparty synchronised rather than silently skipped.

## What did not change

The reply's own framing: a gap fill still carries `MsgSeqNum = BeginSeqNo` and allocates no sender sequence number, because it occupies the range it replaces. The bounds from #25 — tags 7 and 16 validated, `16=0` as infinity, `min(EndSeqNo + 1, next_sender)` — are intact and still tested.

## Two limits, stated rather than left to be discovered

- **The store is opt-in.** Without one attached, the whole range is still answered with a single gap fill, exactly as before.
- **`MemoryStore` is not persistent.** Nothing is replayable after a restart.

Both are in `doc/fix_operations.md`, and the Phase 1 "Resend Request" entry is ticked with them attached rather than ticked bare.

## Tests

Workspace goes to **428 passing**, including an end-to-end test that drives a stub acceptor sending a ResendRequest and asserts the replayed messages carry `43=Y` and `122` with the original sending time — not merely that something was sent.

## Verification

```
cargo test --workspace                                    # 428 passed
```

### DIFF
--- a/ironfix-engine/src/initiator.rs
+++ b/ironfix-engine/src/initiator.rs
@@ -44,10 +44,21 @@
 //! Sequence recovery follows `doc/fix_operations.md`: a `SequenceReset` with
 //! `GapFillFlag` (123) = Y is an ordinary sequenced message and is validated
 //! against its own `MsgSeqNum`, while Reset mode alone may ignore it; an
-//! inbound `ResendRequest` is answered with a `SequenceReset`-GapFill bounded
-//! by `EndSeqNo`. **The engine does not use a message store for replay, so no
-//! message is ever actually replayed** — a resend request is always answered
-//! with a gap fill.
+//! inbound `ResendRequest` is answered within the bound set by `EndSeqNo`.
+//!
+//! # Resend and the message store
+//!
+//! [`Initiator::with_store`] attaches a [`MessageStore`]. Every sequenced
+//! outbound frame is then filed under its `MsgSeqNum` before it goes on the
+//! wire, and an inbound `ResendRequest` replays the stored application
+//! messages with `PossDupFlag` (43) = Y and their original `SendingTime` in
+//! `OrigSendingTime` (122). Administrative messages in the range, and any
+//! sequence number the store cannot produce, are covered by
+//! `SequenceReset`-GapFill, so the counterparty's expectation always lands
+//! exactly where the request asked it to.
+//!
+//! **Without a store nothing can be replayed** and the whole requested range
+//! is answered with one gap fill.
 //!
 //! All sequence arithmetic goes through the checked
 //! [`SequenceManager::try_allocate_sender_seq`] /
@@ -99,6 +110,7 @@ use ironfix_session::{
     Active, Disconnected, HeartbeatManager, LogoutPending, SequenceManager, Session, SessionConfig,
     TestRequestOutcome,
 };
+use ironfix_store::MessageStore;
 use ironfix_transport::FixCodec;
 use std::num::NonZeroU64;
 use std::sync::{Arc, Mutex};
@@ -117,12 +129,20 @@ const DEFAULT_OUTBOUND_CAPACITY: usize = 1024;
 /// Reactor tick granularity for heartbeat/timeout checks.
 const TICK_INTERVAL: Duration = Duration::from_millis(100);
 
+/// Number of stored messages a resend reads from the store per page.
+///
+/// A `ResendRequest` with `EndSeqNo` (16) = 0 asks for the whole session, so the
+/// replay reads in bounded batches and yields between them rather than
+/// allocating and locking over the entire history at once. The value trades the
+/// per-page allocation ceiling against how often the reactor yields; a few
+/// hundred frames keeps both modest.
+const RESEND_PAGE_LIMIT: usize = 256;
+
 /// Client-side FIX engine.
 ///
 /// Owns the session configuration and the [`Application`] callbacks. Each
 /// call to [`Initiator::connect`] establishes one live session and returns
 /// a [`Connection`] handle for it.
-#[derive(Debug)]
 pub struct Initiator<A: Application = NoOpApplication> {
     /// Session configuration.
     config: SessionConfig,
@@ -143,6 +163,25 @@ pub struct Initiator<A: Application = NoOpApplication> {
     initial_sequences: Option<(u64, u64)>,
     /// Capacity of the outbound command queue.
     outbound_capacity: usize,
+    /// Optional message store: outbound frames are filed here and replayed
+    /// from here when the counterparty asks for a resend.
+    store: Option<Arc<dyn MessageStore>>,
+}
+
+/// Written by hand rather than derived: `dyn MessageStore` is not `Debug`, and
+/// requiring it of every implementation to print one field is the wrong trade.
+impl<A: Application> std::fmt::Debug for Initiator<A> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Initiator")
+            .field("config", &self.config)
+            .field("session_id", &self.session_id)
+            .field("version", &self.version)
+            .field("connect_timeout", &self.connect_timeout)
+            .field("initial_sequences", &self.initial_sequences)
+            .field("outbound_capacity", &self.outbound_capacity)
+            .field("store", &self.store.is_some())
+            .finish_non_exhaustive()
+    }
 }
 
 impl<A: Application + 'static> Initiator<A> {
@@ -181,7 +220,34 @@ impl<A: Application + 'static> Initiator<A> {
             connect_timeout: Duration::from_secs(30),
             initial_sequences: None,
             outbound_capacity: DEFAULT_OUTBOUND_CAPACITY,
+            store: None,
         }
+    }
+
+    /// Attaches a message store to the session.
+    ///
+    /// With a store the engine files every sequenced outbound frame under its
+    /// `MsgSeqNum` and, on an inbound `ResendRequest` (35=2), replays the
+    /// stored **application** messages with `PossDupFlag` (43) = Y and the
+    /// original `SendingTime` in `OrigSendingTime` (122). Administrative
+    /// messages, and any sequence number the store cannot produce, are covered
+    /// by `SequenceReset`-GapFill instead — see
+    /// [`Initiator::connect`](Initiator::connect) and
+    /// `doc/fix_operations.md`, "Resend Request".
+    ///
+    /// Without a store the engine has nothing to replay and answers the whole
+    /// requested range with one gap fill.
+    ///
+    /// The store also carries sequence numbers: unless `reset_on_logon` is set
+    /// or [`Initiator::with_initial_sequences`] is called, the session starts
+    /// from the counters the store reports, and both counters are mirrored back
+    /// into it as the session runs. Note that
+    /// [`MemoryStore`](ironfix_store::MemoryStore) is the only implementation
+    /// today and is **not** persistent, so this does not yet survive a restart.
+    #[must_use]
+    pub fn with_store(mut self, store: Arc<dyn MessageStore>) -> Self {
+        self.store = Some(store);
+        self
     }
 
     /// Sets the TCP connect timeout (default 30s).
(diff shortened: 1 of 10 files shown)


ironfix-engine/src/wire.rs
```diff
@@ -55,10 +195,122 @@ pub(crate) fn owned_from_frame(frame: &[u8]) -> Result<OwnedMessage, DecodeError
...
+/// because both change with the two inserted fields.
+///
+/// # Errors
+/// Returns [`ResendError`] if the stored frame does not decode, carries no
+/// `SendingTime` (52), or cannot be re-encoded. The caller must gap-fill the
+/// sequence number instead of skipping it.
+pub(crate) fn resend_frame(stored: &[u8]) -> Result<BytesMut, ResendError> {
+    let raw = decode_frame(stored)?;
+    let begin_string = raw.begin_string()?;
+    let orig_sending_time = raw
```
joaquinbejar: 🔴 **Critical**: a resend-of-a-resend derives OrigSendingTime (122) from the previous transmission SendingTime (52) after discarding the existing 122, so the true original time is lost. The coverage job reproduces this when the timestamps differ. Prefer an existing 122 and fall back to 52 only for the first resend.

ironfix-engine/src/initiator.rs
```diff
@@ -208,28 +357,38 @@ impl<A: Application + 'static> Initiator<A> {
...
         let logon = factory.logon(
             seq,
             self.config.heartbeat_interval_secs(),
             self.config.reset_on_logon,
-        );
+        )?;
         if let Ok(mut owned) = wire::owned_from_frame(&logon) {
             self.application.to_admin(&mut owned, &session_id).await;
         }
+        persist_outbound(
```
joaquinbejar, line 487: 🔴 **Critical**: this stores and sends the Logon after allocating its sequence number, but the store next-sender counter is not mirrored until the reactor starts. If the peer receives the Logon and the handshake then fails, reconnecting from the same store reuses that MsgSeqNum. Persist the advanced counter before any numbered frame can reach the wire, including handshake frames, and test the timeout/reconnect path.

ironfix-engine/src/initiator.rs
```diff
@@ -157,6 +238,55 @@ impl<A: Application + 'static> Initiator<A> {
...
+    ///    messages would leave two different streams filed under the same
+    ///    numbers, and a later `ResendRequest` could not tell them apart.
+    /// 2. [`Initiator::with_initial_sequences`] — an explicit instruction from
+    ///    the consumer outranks whatever the store remembers.
+    /// 3. The store's own counters, refreshed from its backing medium first.
+    /// 4. A fresh session at 1/1.
+    async fn seed_sequences(&self) -> SequenceManager {
+        if self.config.reset_on_logon {
+            if let Some(store) = &self.store
+                && let Err(err) = store.reset().await
```
joaquinbejar, line 309: 🔴 **Critical**: reset and refresh failures are only logged, after which the connection starts at sequence 1 against stale or unknown store state. That can reuse sequence numbers and later replay traffic from the prior stream. Make sequence seeding fallible and abort connect when an attached store cannot reset or refresh.

ironfix-engine/src/initiator.rs
```diff
@@ -348,9 +591,49 @@ impl<A: Application + 'static> Initiator<A> {
...
+    store: Option<&Arc<dyn MessageStore>>,
+    session_id: &SessionId,
+    seq: u64,
+    msg_type: &MsgType,
+    frame: &[u8],
+) {
+    let Some(store) = store else {
+        return;
+    };
+    if let Err(err) = store.store(seq, msg_type, frame).await {
```
joaquinbejar, line 829: 🔴 **Critical**: a failed store write is allowed onto the wire, and the replay path likewise converts unreadable or unrebuildable application messages into GapFills. The repository FIX contract reserves GapFill for administrative traffic; advancing past an unavailable order silently loses business data. Do not send an application frame that was not stored, and fail recovery when stored application traffic cannot be retrieved or rebuilt.

ironfix-store/src/traits.rs
```diff
@@ -20,26 +21,54 @@ use ironfix_core::message::OwnedMessage;
...
-    async fn get_range(&self, begin: u64, end: u64) -> Result<Vec<OwnedMessage>, StoreError>;
+    /// Returns [`StoreError::InvalidRange`] if `begin` is above `end` (with
+    /// `end` = 0 meaning infinity, which is never below `begin`), or `begin`
+    /// is 0, which is not a legal `MsgSeqNum`;
+    /// [`StoreError::RangeNotAvailable`] if the range is legal but holds no
+    /// stored message at all; or another `StoreError` if retrieval fails.
+    ///
+    /// An entirely empty range is an error rather than an empty vector so a
+    /// caller cannot mistake "nothing was stored" for "nothing was asked for".
+    async fn get_range(&self, begin: u64, end: u64) -> Result<Vec<StoredMessage>, StoreError>;
```
joaquinbejar, line 71: 🔴 **Critical**: EndSeqNo=0 lets a peer request the entire history, and this Vec contract materializes every result at once; MemoryStore also builds it while holding a synchronous read lock. A sufficiently long session can therefore suffer unbounded allocation and reactor starvation from one request. Use bounded pages or a streaming contract and yield between batches.

ironfix-store/src/traits.rs
```diff
@@ -20,26 +21,54 @@ use ironfix_core::message::OwnedMessage;
...
-    /// * `seq_num` - The message sequence number
-    /// * `message` - The raw message bytes
+    /// * `seq_num` - The `MsgSeqNum` (34) the message was sent under
+    /// * `msg_type` - The `MsgType` (35) the message was sent as
+    /// * `message` - The complete raw frame bytes
     ///
     /// # Errors
     /// Returns `StoreError` if the message cannot be stored.
-    async fn store(&self, seq_num: u64, message: &[u8]) -> Result<(), StoreError>;
+    async fn store(
```
joaquinbejar, line 39: 🟠 **Warning**: MessageStore::store and get_range are breaking public signature changes; this delta also changes StoreError fields and adds StoredMessage and Initiator::with_store. These changes can fit the planned 0.4.0 compatibility bump, but the PR needs the required complete semver inventory and ADR so downstream migration is explicit.

joaquinbejar (commented):
Risk: critical. Effective verdict: REQUEST_CHANGES. Submitted as COMMENT because this PR is self-authored. The resend/store design is directionally useful, but the exact head has a failing coverage check and can lose original timestamps, reuse sequence numbers, silently gap-fill unavailable business traffic, and materialize unbounded history.

ironfix-engine/src/wire.rs
joaquinbejar: Good catch — fixed in ddc51b6. resend_frame prefers an existing OrigSendingTime(122) now and only falls back to SendingTime(52) on the first resend, so a resend-of-a-resend keeps the true original time instead of overwriting it with the prior replay's. I also made the guard test deterministic (hand-built intermediate frame with distinct 52/122) instead of relying on all sends landing in one millisecond, exactly the coincidence you flagged.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed in ddc51b6 — the advanced sender counter is mirrored into the store right after each pre-reactor handshake frame's seq is allocated and before it reaches the wire (Logon, the pending ResendRequest, reject/logout), so a handshake failure after the Logon can't leave a reused MsgSeqNum behind on a store reused across a reconnect. Added a reconnect test for it.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed in ddc51b6 — seed_sequences is fallible now and aborts connect() with EngineError::Store on a reset (reset_on_logon) or refresh failure, instead of silently falling back to 1/1 against a store that still holds the previous stream. Agreed it's latent with MemoryStore today, but it's exactly the hole a persistent store would fall into, which is what the wiring is for, so fail-closed is the right call.

ironfix-engine/src/initiator.rs
joaquinbejar: I'm going to leave this one as-is. fix_operations.md explicitly lists application messages the store never held or can't rebuild among the things that get gap-filled, so that's the spec-defined degradation rather than an admin-only reservation; and continuing past a failed store write is the documented policy — the message just becomes a hole the resend path gap-fills, instead of tearing down a healthy session over a store hiccup. The operational point is real, but changing it would be a policy change against our own binding spec, not a bug fix.

ironfix-store/src/traits.rs
joaquinbejar: Fixed in ddc51b6 — added a paged get_page(begin, end, limit) to the store and the replay reads in 256-frame batches with a yield between them, so a single ResendRequest with EndSeqNo=0 can't materialize an O(N) Vec or stall the reactor. One clarification: the payload clone was already a Bytes refcount bump rather than a byte copy, so this was the Vec/lock-hold cost, not unbounded byte duplication.

ironfix-store/src/traits.rs
joaquinbejar: Acknowledged — the signature changes are real and allowed at 0.4.0. I'm enumerating the full public delta (store now taking msg_type, the new get_page/StoredMessage, Initiator::with_store, the StoreError fields) in the PR description for the 0.4.0 semver record, with the migration/ADR sign-off going to architect.

PR #44, opened 2026-07-22
fix(engine): make the outbound and callback paths do what they promise

### LINKED ISSUE #11: ironfix-engine: outbound/application path — to_admin/to_app mutations silently discarded, reactor blocks on stalled peers and slow callbacks, admin MsgTypes and reserved tags injectable via Connection::send

joaquinbejar:
## Summary

Makes the outbound and callback paths do what their documentation promises: mutable interception that reaches the wire, a public send that cannot inject administrative traffic, and a reactor a slow callback cannot stall.

Closes #11.

## Stack

- **Base branch:** `stack/17-session-config` (PR #41)

## What was wrong

**`to_admin` / `to_app` mutations were silently discarded.** The contract promised mutable interception — the documented use is stamping authentication fields onto an outbound Logon — but the engine built a **detached** `OwnedMessage`, handed it to the application, and then transmitted the *original* frame. Neither mutating nor replacing it had any effect; the callback was a lie. The callbacks now operate on the `OutboundMessage` the engine actually encodes, so a stamped field reaches the wire.

**Administrative MsgTypes and reserved tags were injectable through `Connection::send`.** The public path accepted any MsgType, so an application could emit Logon, Logout or SequenceReset outside the state machine, and could append reserved header/trailer tags (8, 9, 10, 34, 35, 43, 49, 50, 52, 56, 57, 122, 1128) producing duplicates a conforming counterparty rejects. Both are refused at the public boundary now — administrative messages belong to the session layer alone.

**The reactor blocked on every application callback.** They were awaited inline, so a slow `from_app` stalled socket reads, heartbeat generation and timeout handling, and a stalled peer blocked all three. Inbound application messages go to a **separate dispatcher task** now, so `from_app` cannot stall the reactor. `to_admin`, `to_app` and `from_admin` stay inline by design — each for a documented reason, since each must run at a point in the outbound or handshake sequence where deferring it would change behaviour — and the ordering the session layer depends on is preserved.

## Two deferred follow-ups, now done

Both were explicitly punted to this issue by earlier PRs:

- **`OutboundMessage` is validated before its sequence number is allocated.** Encoding became fallible earlier in the stack, and a value with no legal wire form was detected *after* the number was spent, tearing the session down to avoid a gap. Validating first turns that into a clean rejection with the session intact.
- **`MessageFactory` reuses one encoder** rather than building a fresh one per message.

No performance number is claimed — there is no benchmark harness on this branch.

## What must not regress

The inbound path from #25/#37 is untouched and still tested: GapFill distinguished from Reset, bounded ResendRequest reply carrying `MsgSeqNum = BeginSeqNo`, CompID reason 9, inbound `141=Y` requiring MsgSeqNum 1, checked sequence arithmetic, and the rejected-in-sequence-GapFill-consumes-its-number path.

`to_admin`/`to_app` take `&mut OutboundMessage`; `EngineError` gains `ReservedMsgType` and `ReservedTag`; `RESERVED_TAGS` is exposed. Workspace is at 0.4.0.

## Tests

Workspace goes to **493 passing**, including a stub-acceptor test with a deliberately slow callback that asserts heartbeats still go out — the one that proves the reactor no longer blocks.

## Verification

```
cargo test --workspace                                    # 493 passed
```

### DIFF
--- a/ironfix-engine/src/application.rs
+++ b/ironfix-engine/src/application.rs
@@ -9,8 +9,9 @@
 //! This module defines the callback interface for handling FIX messages,
 //! following the QuickFIX pattern with async support.
 
+use crate::outbound::OutboundMessage;
 use async_trait::async_trait;
-use ironfix_core::message::{OwnedMessage, RawMessage};
+use ironfix_core::message::RawMessage;
 
 /// Session identifier.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
@@ -103,6 +104,29 @@ impl RejectReason {
 ///
 /// Implement this trait to receive callbacks for session events
 /// and message processing.
+///
+/// # Where each callback runs
+///
+/// [`Application::to_admin`] and [`Application::to_app`] run **inline in the
+/// session reactor**, on the message the engine is about to encode and
+/// number. That is what makes their mutations effective, and it is why they
+/// cannot be moved off the reactor: the message they hand back is the one that
+/// gets a sequence number. A slow implementation delays the session's own
+/// timers.
+///
+/// [`Application::from_admin`] also runs inline, because its verdict decides
+/// what the session does next — whether a `SequenceReset` is applied, whether
+/// the Logon is accepted — and that decision cannot be taken before the
+/// verdict exists.
+///
+/// [`Application::from_app`] runs on a **separate dispatcher task** fed by a
+/// bounded queue, so a slow handler no longer stalls socket reads, heartbeat
+/// generation or timeout detection. Messages reach it in ascending
+/// `MsgSeqNum` order. The only consequence of a rejection — a session-level
+/// Reject carrying `RefSeqNum` — is emitted by the reactor when it next polls,
+/// so it may follow messages that arrived after the rejected one. See
+/// [`Initiator::with_app_queue_capacity`](crate::Initiator::with_app_queue_capacity)
+/// for the queue depth and what happens when it fills.
 #[async_trait]
 pub trait Application: Send + Sync {
     /// Called when a session is created.
@@ -123,14 +147,26 @@ pub trait Application: Send + Sync {
     /// * `session_id` - The session identifier
     async fn on_logout(&self, session_id: &SessionId);
 
-    /// Called before sending an admin message.
+    /// Called before an administrative message (Logon, Heartbeat, Logout, …)
+    /// is encoded.
     ///
-    /// Allows modification of outgoing admin messages (Logon, Heartbeat, etc.).
+    /// The body is mutable and the mutation is effective: the engine encodes
+    /// exactly the message this callback hands back. Fields appended here land
+    /// after the ones the session layer wrote, in insertion order. The
+    /// canonical use is stamping `Username` (553) and `Password` (554) onto the
+    /// outbound Logon.
+    ///
+    /// The standard header is not visible: it is stamped after this returns,
+    /// and `MsgSeqNum` (34) is not decided until the frame exists. A body that
+    /// carries a tag the engine stamps itself, or a value with no legal wire
+    /// form, is refused — the message is dropped with a warning and the session
+    /// continues, having spent no sequence number. See
+    /// [`crate::outbound::RESERVED_TAGS`].
     ///
     /// # Arguments
-    /// * `message` - The message to be sent (mutable)
+    /// * `message` - The message about to be encoded (mutable)
     /// * `session_id` - The session identifier
-    async fn to_admin(&self, message: &mut OwnedMessage, session_id: &SessionId);
+    async fn to_admin(&self, message: &mut OutboundMessage, session_id: &SessionId);
 
     /// Called when an admin message is received.
     ///
@@ -147,14 +183,16 @@ pub trait Application: Send + Sync {
         session_id: &SessionId,
     ) -> Result<(), RejectReason>;
 
-    /// Called before sending an application message.
+    /// Called before an application message is encoded.
     ///
-    /// Allows modification of outgoing application messages.
+    /// Same contract as [`Application::to_admin`]: the body is mutable, the
+    /// mutation is effective, and the header — including `MsgSeqNum` (34) — is
+    /// stamped afterwards.
     ///
     /// # Arguments
-    /// * `message` - The message to be sent (mutable)
+    /// * `message` - The message about to be encoded (mutable)
     /// * `session_id` - The session identifier
-    async fn to_app(&self, message: &mut OwnedMessage, session_id: &SessionId);
+    async fn to_app(&self, message: &mut OutboundMessage, session_id: &SessionId);
 
     /// Called when an application message is received.
     ///
@@ -184,7 +222,7 @@ impl Application for NoOpApplication {
 
     async fn on_logout(&self, _session_id: &SessionId) {}
 
-    async fn to_admin(&self, _message: &mut OwnedMessage, _session_id: &SessionId) {}
+    async fn to_admin(&self, _message: &mut OutboundMessage, _session_id: &SessionId) {}
 
     async fn from_admin(
         &self,
@@ -194,7 +232,7 @@ impl Application for NoOpApplication {
         Ok(())
     }
 
-    async fn to_app(&self, _message: &mut OwnedMessage, _session_id: &SessionId) {}
+    async fn to_app(&self, _message: &mut OutboundMessage, _session_id: &SessionId) {}
 
     async fn from_app(
         &self,
--- a/ironfix-engine/src/initiator.rs
+++ b/ironfix-engine/src/initiator.rs
@@ -63,7 +63,28 @@
 //! All sequence arithmetic goes through the checked
 //! [`SequenceManager::try_allocate_sender_seq`] /
 //! [`SequenceManager::try_increment_target_seq`]; an exhausted counter tears
-//! the session down rather than wrapping.
+//! the session down rather than wrapping. A sender number is spent only once
+//! the frame that carries it exists, so a body with no legal wire form is a
+//! dropped message rather than a hole the counterparty must resend over.
+//!
+//! # What blocks the reactor, and what does not
+//!
+//! The reactor is a single task owning the socket. Two things used to be able
+//! to park it indefinitely, and neither can now:
+//!
+//! - **A stalled peer.** A counterparty that stops reading closes its receive
+//!   window and parks a socket write forever. Every write is therefore bounded
+//!   by [`Initiator::with_write_timeout`], and its expiry closes the session —
+//!   the same verdict heartbeat detection would have reached, reached in
+//!   bounded time.
+//! - **A slow `from_app`.** Inbound application messages are handed to a
+//!   separate dispatcher task over a bounded queue
+//!   ([`Initiator::with_app_queue_capacity`]), so an application handler never
+//!   delays a socket read, a heartbeat, or timeout detection.
+//!
+//! `to_admin`, `to_app` and `from_admin` still run inline, because each one's
+//! result decides what the reactor does next. See [`Application`] for the
+//! ordering that buys and the ordering it costs.
 //!
 //! # Example
 //!
@@ -97,8 +118,9 @@
 use crate::application::{Application, NoOpApplication, RejectReason, SessionId};
 use crate::connection::{Command, Connection, SessionRuntime};
 use crate::error::EngineError;
-use crate::wire::{self, MessageFactory, PeerIdentity, UnsupportedVersion};
-use bytes::BytesMut;
+use crate::outbound;
+use crate::wire::{self, MessageFactory, PeerIdentity, PendingMessage, UnsupportedVersion};
+use bytes::{Bytes, BytesMut};
 use futures_util::{SinkExt, StreamExt};
 use ironfix_core::error::{DecodeError, EncodeError};
 use ironfix_core::message::{MsgType, RawMessage};
@@ -117,6 +139,7 @@ use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use tokio::net::{TcpStream, ToSocketAddrs};
 use tokio::sync::{mpsc, watch};
+use tokio::task::JoinHandle;
 use tokio::time::{MissedTickBehavior, interval, timeout};
 use tokio_util::codec::Framed;
 
@@ -125,6 +148,27 @@ type FixFramed = Framed<TcpStream, FixCodec>;
 
 /// Default capacity of the outbound command queue.
 const DEFAULT_OUTBOUND_CAPACITY: usize = 1024;
+
+/// Default capacity of the inbound application queue, in messages.
+///
+/// The queue is bounded by message **count**, not bytes, and each entry retains
+/// its whole frame (a reference-counted [`Bytes`] up to
+/// [`SessionConfig::max_message_size`]). The worst-case retained memory is
+/// therefore `capacity * max_message_size`: with the defaults, 1024 * 1 MiB =
+/// 1 GiB, and up to 1024 * 64 MiB = 64 GiB if `max_message_size` is raised to
+/// its ceiling. See [`Initiator::with_app_queue_capacity`] to size it.
+const DEFAULT_APP_QUEUE_CAPACITY: usize = 1024;
+
+/// Default bound on a single socket write.
+const DEFAULT_WRITE_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// How long the reactor waits for the application dispatcher to finish the
+/// messages already queued before it reports the session closed.
+///
+/// The dispatcher always terminates — its queue is closed when the reactor
+/// drops the sender — so this only bounds how long `on_logout` waits to follow
+/// the last `from_app` rather than run beside it.
+const APP_DRAIN_TIMEOUT: Duration = Duration::from_secs(5);
 
 /// Reactor tick granularity for heartbeat/timeout checks.
 const TICK_INTERVAL: Duration = Duration::from_millis(100);
@@ -159,10 +203,14 @@ pub struct Initiator<A: Application = NoOpApplication> {
     config_check: Result<(), SessionConfigError>,
     /// TCP connect timeout.
     connect_timeout: Duration,
+    /// Bound on a single socket write.
+    write_timeout: Duration,
     /// Initial (sender, target) sequence numbers for session continuity.
     initial_sequences: Option<(u64, u64)>,
     /// Capacity of the outbound command queue.
     outbound_capacity: usize,
+    /// Capacity of the inbound application queue.
+    app_queue_capacity: usize,
     /// Optional message store: outbound frames are filed here and replayed
     /// from here when the counterparty asks for a resend.
     store: Option<Arc<dyn MessageStore>>,
@@ -177,8 +225,10 @@ impl<A: Application> std::fmt::Debug for Initiator<A> {
             .field("session_id", &self.session_id)
             .field("version", &self.version)
             .field("connect_timeout", &self.connect_timeout)
+            .field("write_timeout", &self.write_timeout)
             .field("initial_sequences", &self.initial_sequences)
             .field("outbound_capacity", &self.outbound_capacity)
+            .field("app_queue_capacity", &self.app_queue_capacity)
             .field("store", &self.store.is_some())
             .finish_non_exhaustive()
     }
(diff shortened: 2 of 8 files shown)


ironfix-engine/src/application.rs
```diff
@@ -123,14 +147,26 @@ pub trait Application: Send + Sync {
...
+    /// form, is refused — the message is dropped with a warning and the session
+    /// continues, having spent no sequence number. See
+    /// [`crate::outbound::RESERVED_TAGS`].
     ///
     /// # Arguments
-    /// * `message` - The message to be sent (mutable)
+    /// * `message` - The message about to be encoded (mutable)
     /// * `session_id` - The session identifier
-    async fn to_admin(&self, message: &mut OwnedMessage, session_id: &SessionId);
+    async fn to_admin(&self, message: &mut OutboundMessage, session_id: &SessionId);
```
joaquinbejar, line 169: 🔴 Critical: this makes OutboundMessage the documented password-stamping object, but the type's Debug output includes every raw field value. Logging it after adding tags 554 or 925 leaks credentials. Remove the value-bearing Debug implementation or replace it with a redacting formatter before exposing the message here.

ironfix-engine/src/outbound.rs
```diff
@@ -4,16 +4,128 @@
...
+///
+/// # Errors
+/// [`EngineError::ReservedTag`] for a tag in [`RESERVED_TAGS`], and
+/// [`EngineError::InvalidField`] for tag `0`, an empty value, or a value
+/// carrying the SOH delimiter — which would terminate its own field early and
+/// let the remainder inject further fields into the frame.
+///
+/// The reported reason never quotes the value: an outbound Logon body carries
+/// `Password` (554) and `NewPassword` (925).
+pub(crate) fn check_body(message: &OutboundMessage) -> Result<(), EngineError> {
```
joaquinbejar, line 138: 🔴 Critical: this only validates generic field syntax after to_admin. The callback can still delete or duplicate required session-owned fields such as Logon 98/108, TestRequest 112, ResendRequest 7/16, or GapFill 36, and the engine emits the malformed message. Keep those fields immutable or validate required presence and uniqueness per MsgType after the callback, with malformed-callback tests.

joaquinbejar (commented):
Effective verdict: REQUEST_CHANGES. The new callback and dispatch paths have critical credential, sequence-continuity, and protocol-state risks. Submitted as COMMENT because this is a self-authored pull request; the 0.4.0 breaking-version bump is appropriate, but the callback contract change also needs its architecture record.

ironfix-engine/src/initiator.rs
```diff
@@ -465,20 +1004,124 @@ async fn run_reactor<A: Application + 'static>(
...
+    ///
+    /// # Errors
+    /// Every [`EngineError`] returned here is terminal for the session: an
+    /// exhausted sender counter, a write timeout, or a transport failure.
+    async fn send(
+        &mut self,
+        framed: &mut FixFramed,
+        mut pending: PendingMessage,
+    ) -> Result<(), EngineError> {
+        if !self.prepare(&mut pending).await {
```
joaquinbejar, line 1471: 🔴 Critical: a callback-rejected administrative message returns the same success as a completed write. Callers can then arm a TestRequest timeout, enter LogoutPending, or mark a ResendRequest pending although nothing reached the peer. Return an explicit Sent/Dropped outcome and advance protocol state only after Sent; cover all three call paths.

ironfix-engine/src/initiator.rs
```diff
@@ -465,20 +1004,124 @@ async fn run_reactor<A: Application + 'static>(
...
+                tracing::warn!(
+                    session = %self.session_id,
+                    error = %err,
+                    "dropping outbound message with no legal wire form; no sequence number spent"
+                );
+                return Ok(());
+            }
+            Err(WriteFailure::Fatal(err)) => return Err(err),
+        }
+        let allocated = self.runtime.sequences.try_allocate_sender_seq()?.value();
```
joaquinbejar, line 1487: 🔴 Critical: the frame has already been written when the sender sequence is allocated. At exhaustion, after an ambiguous write, or if the task is cancelled between these operations, the peer may receive a sequence number that local state can reuse. Encode first, allocate and persist the number, then write; add an active-session exhaustion/failed-write continuity test.

ironfix-engine/src/initiator.rs
```diff
@@ -603,155 +1549,168 @@ impl<A: Application> Reactor<A> {
...
+        seq: u64,
+        frame: &Bytes,
+    ) -> Result<Phase, SessionClosed> {
+        let msg_type = raw.msg_type();
+        if !msg_type.is_admin() {
+            let queued = AppFrame {
+                frame: frame.clone(),
+                seq,
+            };
+            return match self.app_tx.try_send(queued) {
```
joaquinbejar: 🔴 Critical: queue exhaustion is discovered only after the inbound sequence expectation has advanced. The application never receives this business message, while reconnect continuity records it as consumed and cannot request it again. Reserve queue capacity before consuming MsgSeqNum, or leave an unqueued frame as the next expected sequence, and test queue-full reconnect continuity.

ironfix-engine/src/initiator.rs
```diff
@@ -453,6 +981,17 @@ async fn run_reactor<A: Application + 'static>(
         }
     };
 
+    // Closing the queue is what stops the dispatcher; joining it is what keeps
+    // `on_logout` from running beside a `from_app` still in flight. A wedged
+    // handler bounds the wait rather than the shutdown.
+    drop(ctx.app_tx);
+    if timeout(APP_DRAIN_TIMEOUT, dispatcher).await.is_err() {
```
joaquinbejar: 🟠 Warning: timing out this moved JoinHandle detaches the dispatcher instead of cancelling it, so a wedged from_app can continue after on_logout and wait_closed. Await a mutable handle, close the rejection receiver, then abort and await the dispatcher on expiry; cover a callback exceeding the drain bound.

ironfix-engine/src/initiator.rs
```diff
@@ -72,6 +135,20 @@ type FixFramed = Framed<TcpStream, FixCodec>;
 /// Default capacity of the outbound command queue.
 const DEFAULT_OUTBOUND_CAPACITY: usize = 1024;
 
+/// Default capacity of the inbound application queue.
+const DEFAULT_APP_QUEUE_CAPACITY: usize = 1024;
```
joaquinbejar, line 160: 🟠 Warning: this bounds message count but not retained bytes. With the default 1 MiB frame ceiling, a slow callback can retain about 1 GiB per session, and the supported 64 MiB ceiling makes the theoretical bound much larger. Add an aggregate byte budget or derive and cap queue depth from max_message_size.

joaquinbejar (commented):
Risk is high. The callback redesign fixes the discarded-mutation bug, but sender sequence state can diverge from the peer and the new dispatcher has shutdown and hot-path regressions. Effective verdict REQUEST_CHANGES. Submitted as COMMENT because this PR is self-authored.

ironfix-engine/src/initiator.rs
joaquinbejar, line 1487: 🔴 **Critical**: this commits `MsgSeqNum` only after `framed.send` completed. A timeout or transport error may happen after the peer accepted some or all bytes, leaving it at sequence N while local state can reuse N; at `u64::MAX` the frame is sent and allocation then fails. Encode under the peeked number, allocate and verify it before writing, and keep it spent on an ambiguous I/O failure.

ironfix-engine/src/initiator.rs
```diff
@@ -465,20 +1004,124 @@ async fn run_reactor<A: Application + 'static>(
...
+    /// # Errors
+    /// Every [`EngineError`] returned here is terminal for the session: an
+    /// exhausted sender counter, a write timeout, or a transport failure.
+    async fn send(
+        &mut self,
+        framed: &mut FixFramed,
+        mut pending: PendingMessage,
+    ) -> Result<(), EngineError> {
+        if !self.prepare(&mut pending).await {
+            return Ok(());
```
joaquinbejar: 🟠 **Warning**: returning `Ok` here makes a callback-invalid administrative message indistinguishable from a successful send. Logout can enter `LogoutPending`, or TestRequest and Resend bookkeeping can advance, although the required frame never left. Make administrative drops terminal or return an explicit sent/dropped outcome, keeping nonfatal drops only for application messages.

ironfix-engine/src/initiator.rs
joaquinbejar: 🟠 **Warning**: timing out while owning this `JoinHandle` drops and detaches the dispatcher; it does not stop it. A wedged `from_app` can outlive session closure and race `on_logout`. Poll a mutable handle, then abort and await it on timeout so every spawned task has a real shutdown path.

ironfix-engine/src/initiator.rs
```diff
@@ -391,10 +793,100 @@ fn teardown(phase: Phase) {
...
+    application: Arc<A>,
+    session_id: SessionId,
+    mut frames: mpsc::Receiver<AppFrame>,
+    rejections: mpsc::Sender<AppRejection>,
+) {
+    while let Some(AppFrame { frame, seq }) = frames.recv().await {
+        // The reactor decoded this frame already; a failure here cannot
+        // happen, and inventing a session action for it is not this task's
+        // job.
+        let Ok(raw) = wire::decode_frame(&frame) else {
```
joaquinbejar, line 1176: 🟠 **Warning**: the reactor already decoded this application frame, but the dispatcher performs the full field scan again and constructs another `SmallVec`; messages over 32 fields can allocate twice. Carry the validated frame plus parsed offsets across the queue so `from_app` can borrow it without reparsing.

ironfix-engine/src/application.rs
joaquinbejar: Good catch on the footgun — fixed in 28efac8. OutboundMessage now has a hand-written Debug that redacts the values of Password (554), NewPassword (925) and RawData (96), so a stray `debug!(?message)` on the very object a to_admin callback stamps a password onto can't leak the secret.

ironfix-engine/src/outbound.rs
joaquinbejar: Fixed in 28efac8 — check_body now refuses an administrative message that lost a required body field, so a to_admin that removes HeartBtInt(108) from a Logon or TestReqID(112) from a TestRequest gets EngineError::MissingRequiredField instead of the malformed frame reaching the wire. It covers Logon 98/108, TestRequest 112, ResendRequest 7/16 and SequenceReset 36.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed in 28efac8 — send() now returns a Sent::Yes/Dropped outcome instead of a bare Ok, and the callers only advance protocol state on Yes. A Logout dropped by to_admin keeps the session Active, a dropped TestRequest doesn't arm the timeout, and a dropped ResendRequest leaves the gap unmarked so the next frame retries.

ironfix-engine/src/initiator.rs
joaquinbejar: This one stands — #44 already peeks next_sender_seq to frame the message and only spends the number once the write lands (write_at, then try_allocate with an assert that allocated==N), so a wrong MsgSeqNum can't reach the wire. The 'reuse N' path is unreachable: a fatal write failure tears the session down and the engine has no store to reconnect from, so there's no continuity that could reuse the number.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed in 28efac8 — the app frame is now handed to the dispatcher (enqueue_app) before try_increment_target_seq advances the inbound expectation, so a full queue is a clean session close with the number unspent and the peer free to resend, never a message dropped past a number already consumed.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed in 28efac8 — the drain now times out on `&mut dispatcher` (keeping ownership of the handle) and calls dispatcher.abort() on expiry, so a wedged from_app is actually killed instead of detached to outlive on_logout.

ironfix-engine/src/initiator.rs
joaquinbejar: Addressed in 28efac8, as documentation rather than a byte budget — the queue-capacity doc now spells out that the bound is on message count and the worst-case retained memory is capacity * max_message_size (1 GiB at the defaults, up to 64 GiB at the 64 MiB ceiling), so it's raised deliberately with that product in mind. An aggregate byte budget is the natural follow-up if we want a hard memory ceiling.

ironfix-engine/src/initiator.rs
joaquinbejar: Same as the peek-before-allocate thread above — write_at frames under the peeked N and the number is only spent after the write lands, and the reuse path is unreachable in a storeless engine where a write failure is terminal. Stands.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed in 28efac8 — send() returns Sent::Yes/Dropped now, and the Logout/TestRequest/ResendRequest bookkeeping only advances on Yes, so a callback-invalidated admin frame that never left the engine no longer arms follow-on state as if it had been sent.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed in 28efac8 — same dispatcher fix: timeout on `&mut dispatcher` then abort() on expiry, so a wedged from_app can't outlive session close.

ironfix-engine/src/initiator.rs
joaquinbejar: You're right that run_app_dispatcher re-decodes a frame the reactor already parsed, but that's a latency/allocation concern rather than a spec-conformance defect, so I'm leaving it for the perf pass rather than folding it into this branch.

PR #45, opened 2026-07-22
fix(examples): make the examples correct templates, not broken ones

### LINKED ISSUE #19: ironfix-example / Docker: examples are protocol-incorrect templates — hardcoded 34=1, wrong-version fields, f64 prices, overflowable framing, fabricated pmap — plus ~1,500 lines of drifted copy-paste and a non-functional FAST Docker image
## Overview

The 21 runnable examples are the templates users copy, and they teach wrong FIX: eight of nine servers stamp MsgSeqNum=1 on every message, the FIX 4.0 pair sends 4.1+ fields, prices are f64 against the workspace-wide Decimal rule, the shared framing helper repeats the exact BodyLength-overflow bug class fixed in FixCodec (with no tests, and tests inside examples/ would never run), and the FAST pair fabricates its presence map. The shipped fast.Dockerfile sets FAST_HOST while the server reads FIX_HOST, so the container binds loopback and is unreachable. Nine copy-paste pairs have already drifted; dedup plus the fixes belong together.

## Context

Crates / areas touched: `ironfix-example`

### 1. try_decode_message: unchecked BodyLength arithmetic + uncapped buffers (finding 109)
`ironfix-example/examples/common/mod.rs:133` (130-133), `fix40_client.rs:94-101`

`let total = i + j + 1 + len + 7;` with unbounded len from tag 9: 9=usize::MAX panics in debug, wraps to a tiny total in release; read loops have no buffer cap so a peer never sending valid 9= grows BytesMut without bound. Same bug class commit 1110a14 fixed in FixCodec, still in the helper shared by all 18 tag=value examples.

**Fix:** ceiling on len (e.g. 1 MiB), checked_add, cap the accumulation buffer — or reuse ironfix_transport::FixCodec for framing.

### 2. Eight of nine servers hardcode MsgSeqNum(34)=1 (finding 110)
`ironfix-example/examples/fix41_server.rs:88` (counter incremented-but-unused at :75; same in fix40/42/43/50/50sp1/50sp2 and fix44_server_channel)

`e.put_str(34, "1")` on every response — a protocol violation any compliant counterparty (including IronFix's own gap detection) flags. fix44_server.rs does it correctly, so the copies have drifted.

**Fix:** thread the counter into every build_* call as fix44_server does, or hoist the seq-managed response path into examples/common.

### 3. FIX 4.0 pair sends FIX 4.1+ fields and omits 4.0-required ER fields (finding 111)
`ironfix-example/examples/fix40_server.rs:158` (158-162)

The fix40 ExecutionReport puts ExecType(150)/LeavesQty(151) under 8=FIX.4.0 and omits 38/32/31 — a FIX 4.1 message wearing a 4.0 header; the 5.0/SP1/SP2 pairs differ only in constants and exercise nothing SP-specific. Conflicts with 'version support is honest' — examples are the only evidence backing the version matrix.

**Fix:** version-correct the 4.0 ER (drop 150/151, add 38/32/31); for the 5.0 family exercise one SP-specific element or document FIXT-transport-only.

### 4. f64 for prices across order builders and FAST market data (finding 112)
`ironfix-example/examples/fast_server.rs:134`, `fix40_client.rs:143` (+7 siblings), `fast_client.rs:128`

`(price * 100.0) as u64` truncating conversion; eight clients pass `px: f64` formatted for tag 44. CLAUDE.md override 3: f64 for any monetary quantity anywhere is red (the anyhow carve-out does not extend here). fix44_engine_client and fast_server_spsc show the compliant patterns.

**Fix:** Decimal or pre-formatted &str in client builders; integer scaled cents end-to-end for the FAST pair.

### 5. fast_server fabricates the presence map (finding 115)
`ironfix-example/examples/fast_server.rs:118` (118-119; same pattern fast_server_spsc.rs:280-281)

`let pmap_byte: u8 = 0b1111_1100 | 0x80;` is a no-op setting 5 (not the commented 6) bits, then encode_uint stop-bit-varints 0xFC into TWO bytes — the wire carries a uint, not a pmap; fast_client decodes-and-discards it, so the pair interoperates only with itself. PresenceMap/PresenceMapBuilder are exported and used by no example.

**Fix:** build with PresenceMapBuilder and emit raw pmap bytes; parse with PresenceMap in the client.

### 6. fast_client stalls forever with unbounded buffer growth on a corrupt stream (finding 116)
`ironfix-example/examples/fast_client.rs:93` (93-98)

Every decode error is treated as 'not enough data' (`Err(_) => { None }`) with the whole uncapped `Vec::new()` buffer kept: one corrupt message means no bytes are ever consumed while reads append indefinitely.

**Fix:** distinguish UnexpectedEof from other FastError variants (drop/resync on the latter); cap the buffer and disconnect past it.

### 7. fix44_engine_client hardcodes a stale TransactTime (finding 117)
`ironfix-example/examples/fix44_engine_client.rs:103`

`.push_str(60, "20260714-00:00:00.000")` — a fixed literal date, rejected by any TransactTime-validating counterparty; every hand-rolled client uses format_timestamp().

**Fix:** use common::format_timestamp() for tag 60.

### 8. fast.Dockerfile sets FAST_HOST but the server reads FIX_HOST — container unreachable (finding 108)
`Docker/fast.Dockerfile:21`, `common/mod.rs:48`

`ENV FAST_HOST=0.0.0.0` vs `env::var("FIX_HOST")` with DEFAULT_HOST=127.0.0.1 — the container binds 127.0.0.1:9890 and `docker run -p 9890:9890` accepts no external connections; the shipped FAST image is non-functional.

**Fix:** ENV FIX_HOST=0.0.0.0 (or honor FAST_HOST in ExampleConfig); add a docker smoke test connecting to the published port.

### 9. ~1,500 lines of copy-paste across 9 client/server pairs (finding 113)
`ironfix-example/examples/fix42_client.rs:1`

diff-verified: fix50 vs fix50sp1 clients differ at lines 1, 14-15, 26, 51 only; fix41/42/43 build_msg byte-identical; every file re-implements read_msg, builders, and port handling duplicating ExampleConfig. Nine copies means nine edits per fix — and they have already drifted (34=1 bug, fix40 field sets).

**Fix:** extract version-parameterized run_client/run_server + read_msg into examples/common; each fixNN file collapses to constants + a call.

### 10. Shared helpers have zero tests — and tests inside examples/ would never run (finding 114)
`ironfix-example/examples/common/mod.rs:124`

try_decode_message (an untrusted-input parser) and the hand-rolled format_timestamp calendar arithmetic have no tests anywhere; cargo test does not execute example-target tests. The overflow bug above would have been caught by the first malformed-input test.

**Fix:** move helpers to ironfix-example/src (pub mod util) and cover partial/garbage/overflow frames plus format_timestamp vs chrono on leap-year dates.

### 11. No .dockerignore; rust-toolchain 'stable' defeats the image pin (finding 119)
`Docker/fix44.Dockerfile:8` (all 9 Dockerfiles)

`COPY . .` with no .dockerignore — multi-GB context (target/, .git), no dependency-layer caching; rustup inside rust:1.92.0-alpine3.23 resolves the repo's `stable` channel to current latest, downloading a different toolchain mid-build.

**Fix:** add .dockerignore (target/, .git/, Docker/, coverage/); manifests-first COPY for a cached cargo fetch layer; set RUSTUP_TOOLCHAIN=1.92.0.

### 12. fast_server_spsc overclaims SPSC; shutdown flag dead (finding 120)
`ironfix-example/examples/fast_server_spsc.rs:24`

Docs promise 'SPSC lock-free channel' but the code uses crossbeam_channel::bounded (MPMC); `running: Arc<AtomicBool>` is never stored false so both shutdown paths are unreachable — in the example most likely to be copied for hot-path architecture.

**Fix:** use a true SPSC queue (crossbeam-queue ArrayQueue) or reword; wire running.store(false) to ctrl_c.

### 13. fast_server_spsc blocks a tokio worker with crossbeam recv_timeout in async fn; deep-clones per client (finding 126)
`ironfix-example/examples/fast_server_spsc.rs:203` (202-210)

`rx.recv_timeout(Duration::from_millis(10))` inside `async fn market_data_broadcaster` parks a runtime worker up to 10ms per iteration — directly under a comment claiming 'Non-blocking receive with timeout' — and `tx.try_send(encoded.clone())` deep-clones the encoded Vec<u8> per connected client per tick. The exact blocking-in-async antipattern the workspace rules warn against. [source: codex-audit]

**Fix:** dedicated std::thread for the broadcaster (or bridge via tokio::sync); wrap the tick in bytes::Bytes/Arc<[u8]> for refcount fan-out; fix the comment.

## Tasks

- [ ] try_decode_message: unchecked BodyLength arithmetic + uncapped buffers (finding 109)
- [ ] Eight of nine servers hardcode MsgSeqNum(34)=1 (finding 110)
- [ ] FIX 4.0 pair sends FIX 4.1+ fields and omits 4.0-required ER fields (finding 111)
- [ ] f64 for prices across order builders and FAST market data (finding 112)
- [ ] fast_server fabricates the presence map (finding 115)
- [ ] fast_client stalls forever with unbounded buffer growth on a corrupt stream (finding 116)
- [ ] fix44_engine_client hardcodes a stale TransactTime (finding 117)
- [ ] fast.Dockerfile sets FAST_HOST but the server reads FIX_HOST — container unreachable (finding 108)
- [ ] ~1,500 lines of copy-paste across 9 client/server pairs (finding 113)
- [ ] Shared helpers have zero tests — and tests inside examples/ would never run (finding 114)
- [ ] No .dockerignore; rust-toolchain 'stable' defeats the image pin (finding 119)
- [ ] fast_server_spsc overclaims SPSC; shutdown flag dead (finding 120)
- [ ] fast_server_spsc blocks a tokio worker with crossbeam recv_timeout in async fn; deep-clones per client (finding 126)

## Acceptance criteria

- [ ] Framing/timestamp helpers live in ironfix-example/src with tests covering partial frames, garbage, a 9=usize::MAX overflow attempt (bounded, checked), and format_timestamp vs chrono on leap-year dates
- [ ] All server examples stamp a real incrementing MsgSeqNum; a client↔server run shows no gap detection against IronFix's own Initiator
- [ ] fix40 ExecutionReport contains no 150/151 and includes 38/32/31; the 5.0 family either exercises an SP-specific element or documents FIXT-transport-only
- [ ] No f64 monetary value remains in any example (Decimal, pre-formatted &str, or scaled integers)
- [ ] The FAST pair round-trips a real PresenceMap built with PresenceMapBuilder; fast_client distinguishes UnexpectedEof from corrupt data and caps its buffer
- [ ] fix44_engine_client stamps tag 60 from format_timestamp()
- [ ] docker run of the fast image accepts an external connection on the published port (smoke test); .dockerignore exists and the toolchain used in-image is pinned
- [ ] The nine fixNN pairs are collapsed onto shared parameterized run_client/run_server helpers; per-version files contain only constants + a call
- [ ] fast_server_spsc no longer blocks a tokio worker with crossbeam recv_timeout, fans out via cheap-clone Bytes, and its SPSC/shutdown claims match the code

joaquinbejar:
## Summary

Turns the examples from broken templates into correct ones. These are the files a user copies to write their own FIX client, so a wrong one teaches the wrong thing to everyone who reads it — correctness here propagates.

Closes #19.

## Stack

- **Base branch:** `stack/15-store` (PR #42)

## What was wrong, and fixed

- **Hand-rolled framing** with unchecked arithmetic and no size ceiling. The tag=value examples frame through `ironfix-transport::FixCodec` now — the real codec, so no parallel implementation can drift from the one under test.
- **Hard-coded `MsgSeqNum = 1`** in the FIXT servers, demonstrating a session that cannot work. A shared `DemoSession` owns a real `SequenceManager`; IronFix's own `Initiator` runs a full Logon → ExecutionReport → Logout against the demo server with zero gap warnings on either side.
- **`f64` for prices** — the one habit an example must never teach for a monetary value. Now `rust_decimal::Decimal` at the typed seam; FAST prices are scaled integers on the wire that convert to `Decimal`.
- **Wrong-version fields**: the execution report is parameterised by FIX version (4.0 uses `ExecTransType`, 4.3+ uses `ExecType`/`OrdStatus`), unit-tested per version.
- **A fabricated presence map** in the FAST examples, now built and parsed for real.
- **The FAST client stalled** on a decode error; it triages now, buffering only on end-of-input and disconnecting on a corrupt stream.
- **The lock-free FAST server** — renamed from the "SPSC" name it never matched — blocked a Tokio worker on a synchronous receive and deep-cloned per client. It uses a dedicated thread and clones a refcount (`bytes::Bytes`).
- **The FAST Docker image was unreachable**: it set a variable nothing reads, so the server bound to loopback. It reads the right variables now, and the image was built and exercised on its published port.

## Deduplication

~1,500 lines of drifted copy-paste across nine near-identical client/server pairs collapse into shared helpers. **Where it stopped:** each example keeps its own `main`, constants and version-specific docs, so a reader landing on a single file still sees the whole story; the shared session logic and message layouts move into the crate's `src`, which also means the **test runner now exercises them** — 26 library tests where there were none.

## Ran, not just compiled

Every example target was **run**, not merely built: all eight tag=value client/server pairs through a full Logon/order/ExecutionReport/TestRequest/Heartbeat/Logout, the engine `Initiator` against the demo server, both FAST servers against the client, the corrupt-stream disconnect, and the Docker FAST image on its published port. An example that compiles while demonstrating a broken session is exactly the defect this issue is about, so compiling was never the bar.

## Scope notes

`anyhow` stays confined to `examples/` — the documented workspace exception for demo binaries. No library crate changed; the DAG is untouched. The manifest gains `rust_decimal`, `thiserror`, `tokio-util`, `futures-util`, `parking_lot`, all already workspace dependencies — no new external dependency. The top-level README quick start is left to the docs PR (#33) to avoid a conflict.

## Tests

Workspace at **452 passing**.

## Verification

```
cargo build --release --workspace --examples              # zero warnings
cargo test --workspace                                    # 452 passed
```


ironfix-example/examples/common/mod.rs
```diff
@@ -73,67 +182,309 @@ pub fn init_logging() {
...
+                expected,
+                received, "sequence gap: a real acceptor answers this with a ResendRequest"
+            );
+            session.sequences().set_target_seq(received);
+        }
+        SequenceResult::TooLow { expected, received } => {
+            let poss_dup = raw.get_field_str(43) == Some("Y");
+            if poss_dup {
+                info!(expected, received, "possible duplicate (43=Y), ignoring");
+                return Ok(());
```
joaquinbejar: 🔴 Critical: this returns success for a duplicate, so serve_connection continues into MsgType dispatch and executes a retransmitted NewOrderSingle again. Return a distinct Duplicate outcome and skip business handling for it; add a 43=Y duplicate-order regression test.

ironfix-example/src/fast_market_data.rs
```diff
@@ -0,0 +1,367 @@
...
+///
+/// # Arguments
+/// * `message` - The update to encode
+///
+/// # Errors
+/// [`FastError::InvalidString`] if a string field is not representable as FAST
+/// ASCII, or [`FastError::PresenceMapTooLarge`] if the map exceeds the encoder's
+/// ceiling — neither can happen for template 1, but neither is assumed away.
+pub fn encode(message: &MarketData) -> Result<Vec<u8>, FastError> {
+    let pmap = PresenceMapBuilder::new()
```
joaquinbejar, line 125: 🔴 Critical: this is not a conforming FAST 1.1 template header. PMAP bit 0 signals the Template ID copy operator, Template ID uses nullable uInt encoding, and optional fields without operators are represented by nullable encodings rather than PMAP bits. Define the actual operator layout, fix both codecs, and add an independent byte fixture instead of relying only on a self-round-trip.

ironfix-example/src/demo.rs
```diff
@@ -0,0 +1,777 @@
...
+        }
+        if self.version != FixVersion::Fix40 {
+            self.encoder.put_str(150, "0");
+        }
+
+        self.encoder.put_str(39, "0");
+        self.encoder.put_str(55, order.symbol);
+        self.encoder.put_char(54, order.side.as_char());
+
+        if self.version == FixVersion::Fix40 {
```
joaquinbejar: 🔴 Critical: FIX 4.1 takes the else branch and omits required OrderQty(38), LastShares(32), and LastPx(31); its schema requires those in addition to LeavesQty(151). Emit the 4.1 field set and add an explicit dictionary-backed FIX 4.1 fixture test.

ironfix-example/src/demo.rs
```diff
@@ -0,0 +1,777 @@
...
+    pub fn new_order_single(&mut self, order: &DemoOrder<'_>) -> Result<&[u8], DemoError> {
+        self.start(MsgType::NewOrderSingle.as_str())?;
+        if let Some(appl_ver_id) = self.appl_ver_id_for_app_message() {
+            self.encoder.put_str(1128, appl_ver_id);
+        }
+        self.encoder.put_str(11, order.cl_ord_id);
+        self.encoder.put_str(21, "1");
+        self.encoder.put_str(55, order.symbol);
+        self.encoder.put_char(54, order.side.as_char());
+        self.encoder
```
joaquinbejar: 🔴 Critical: this emits TransactTime(60) for every NewOrderSingle, but the FIX 4.0 and FIX 4.1 schemas do not define tag 60 there; it becomes required in FIX 4.2. Gate it by version and validate each generated frame against that version's schema.

ironfix-example/src/demo.rs
```diff
@@ -0,0 +1,777 @@
...
+    /// Builds a `NewOrderSingle` (35=D) for a limit order.
+    ///
+    /// # Arguments
+    /// * `order` - The order to send
+    ///
+    /// # Errors
+    /// As [`DemoSession::heartbeat`].
+    pub fn new_order_single(&mut self, order: &DemoOrder<'_>) -> Result<&[u8], DemoError> {
+        self.start(MsgType::NewOrderSingle.as_str())?;
+        if let Some(appl_ver_id) = self.appl_ver_id_for_app_message() {
```
joaquinbejar: 🔴 Critical: start() has already emitted SenderCompID, TargetCompID, MsgSeqNum, and SendingTime when this appends ApplVerID(1128). FIXT.1.1 places 1128 immediately after MsgType, so strict peers can reject both this frame and the ExecutionReport path at line 388. Stamp the application version inside header construction and test field order.

ironfix-example/src/demo.rs
```diff
@@ -0,0 +1,777 @@
...
+    /// * `order` - The decoded incoming message
+    ///
+    /// # Returns
+    /// `None` if a required field is missing or unusable — an absent
+    /// `ClOrdID`, a `Side` outside the tag-54 enumeration, or a non-numeric
+    /// `OrderQty`. The caller answers that with a session-level `Reject`
+    /// ([`DemoSession::reject`]) rather than substituting a default, which
+    /// would acknowledge an order the client never placed.
+    #[must_use]
+    pub fn from_new_order(order: &ironfix_core::RawMessage<'a>) -> Option<Self> {
```
joaquinbejar, line 143: 🔴 Critical: this accepts and acknowledges orders after checking only 11, 55, 54, and 38. A NewOrderSingle missing HandlInst(21), OrdType(40), version-required TransactTime(60), or conditional limit Price(44) still receives ExecutionReport(New). Make validation version-aware and test every required and conditional field.

ironfix-example/examples/common/mod.rs
```diff
@@ -73,67 +182,309 @@ pub fn init_logging() {
...
+        let inbound_seq = raw.get_field_str(34).and_then(|s| s.parse().ok());
+        info!(msg_type = %raw.msg_type().as_str(), "received");
+
+        match raw.msg_type() {
+            MsgType::Logon => {
+                let response = session.logon()?;
+                send(&mut connection, response).await?;
+            }
+            MsgType::TestRequest => {
+                let test_req_id = raw.get_field_str(112);
```
joaquinbejar, line 493: 🔴 Critical: a TestRequest without required TestReqID(112) is accepted and answered with a Heartbeat that also lacks the required echo. Reject the malformed request with RefTagID 112, and add valid, empty, and missing-ID tests.

ironfix-example/src/demo.rs
```diff
@@ -0,0 +1,777 @@
...
+        self.encoder.put_str(tag, &text);
+    }
+
+    /// Starts a message: clears the buffer and writes the standard header
+    /// fields the encoder does not stamp itself.
+    ///
+    /// `BeginString` (8), `BodyLength` (9) and `CheckSum` (10) belong to
+    /// [`Encoder::finish`]; `MsgType` (35) must be the first body field.
+    fn start(&mut self, msg_type: &str) -> Result<(), DemoError> {
+        let seq = self.sequences.try_allocate_sender_seq()?;
```
joaquinbejar: 🔴 Critical: sender sequence is allocated before the remaining encoding and finish operations can fail, so a rejected body burns a number and the next valid frame creates a gap. Peek the number, encode and finish successfully, then allocate and verify it immediately before returning the frame; test an encoding failure followed by a valid send.

ironfix-example/examples/fast_server.rs
```diff
@@ -1,141 +1,137 @@
...
+
+        info!(
+            seq_num,
+            symbol,
+            price = %Decimal::new(i64::try_from(*price).unwrap_or(i64::MAX), PRICE_SCALE),
+            size = ?message.size,
+            "published"
+        );
+
+        seq_num = seq_num.checked_add(1).unwrap_or(1);
```
joaquinbejar: 🔴 Critical: checked overflow is immediately converted into a silent reset to 1, reusing a market-data sequence number instead of surfacing exhaustion. Stop publishing and propagate a typed error; the same fallback must be fixed in fast_server_lockfree.rs:204.

ironfix-example/src/lib.rs
```diff
@@ -47,6 +60,12 @@
 //! - [`transport`]: Network transport layer
 //! - [`fast`]: FAST protocol encoding and decoding
 //! - [`engine`]: High-level engine facade
+//! - [`demo`]: Version-parameterised message builders backing the runnable
+//!   examples
+//! - [`fast_market_data`]: The FAST template the `fast_*` examples exchange
+
+pub mod demo;
```
joaquinbejar, line 118: 🟠 Warning: these public modules add a large API surface to the published ironfix-example crate, contradicting the PR body's claim that no library crate changed. Document the additive semver event and keep the surface intentionally public or narrow it; this crate is excluded from semver CI.

joaquinbejar (commented):
Effective verdict: REQUEST_CHANGES. The examples currently teach duplicate business processing, invalid versioned FIX layouts, a nonconforming FAST template, and unsafe sequence exhaustion behavior. Submitted as COMMENT because this is a self-authored pull request; the failing coverage check is inherited and already has a current finding on its originating stack PR.

ironfix-example/examples/common/mod.rs
```diff
@@ -73,67 +182,309 @@ pub fn init_logging() {
...
+/// A gap is reported and tolerated: these demos have no message store, so they
+/// cannot answer a `ResendRequest` and must not pretend otherwise.
+fn accept_inbound(session: &DemoSession, raw: &RawMessage<'_>) -> anyhow::Result<()> {
+    let received: u64 = raw
+        .get_field_str(34)
+        .context("inbound message has no MsgSeqNum (34)")?
+        .parse()
+        .context("MsgSeqNum (34) is not a number")?;
+
+    match session.sequences().validate_incoming(received) {
```
joaquinbejar, line 234: 🔴 **Critical**: both non-Ok sequence outcomes still become success for the caller. The gap arm jumps the target counter and dispatches the later frame while silently losing every missing message, and the PossDup arm returns `Ok` so callers process the duplicate too. Return distinct accepted, dropped, and gap outcomes, issue a ResendRequest or close on a gap, and prove neither path reaches order handling.

ironfix-example/src/demo.rs
joaquinbejar: 🔴 **Critical**: this restricts `OrderQty(38)`, `LastShares(32)`, and `LastPx(31)` to FIX 4.0, but FIX 4.1 ExecutionReport still requires all three alongside `LeavesQty(151)`. The fix41 server therefore emits an incomplete message while this PR advertises version-correct templates. Add the FIX 4.1 layout and a recorded fixture checked against its schema.

ironfix-example/src/fast_market_data.rs
```diff
@@ -0,0 +1,367 @@
...
+        .bit(message.sending_time.is_some())
```
joaquinbejar, line 126: 🔴 **Critical**: FAST 1.1 uses the first presence-map bit for TemplateID copy semantics, but this assigns it to `SendingTime` while always encoding TemplateID. The pair self-round-trips because both sides share the same shift, but a conforming peer will read every optional field from the wrong bit. Reserve bit zero for TemplateID and add an interoperability fixture.

ironfix-example/src/lib.rs
joaquinbejar, line 118: 🟠 **Warning**: the PR says no library crate changed, but these public modules add a substantial API to the published `ironfix-example` crate, which is outside the semver workflow. Inventory the new public surface and migration impact explicitly, or keep these example helpers crate-private.

joaquinbejar (commented):
Risk is high. The examples are much clearer, but the session helper can process lost or duplicate orders, FIX 4.1 and FAST frames are not interoperable, and the current coverage job is red. Effective verdict REQUEST_CHANGES. Submitted as COMMENT because this PR is self-authored.

ironfix-example/examples/common/mod.rs
joaquinbejar: Fixed in 85e62d0 — accept_inbound now returns a distinct Inbound::Duplicate for the TooLow+PossDup(43=Y) case, and serve_connection `continue`s on it, so the retransmitted NewOrderSingle is dropped instead of dispatched into a second ExecutionReport. The log line no longer lies.

ironfix-example/src/fast_market_data.rs
joaquinbejar: Fixed in 85e62d0 with the doc-honesty option (b) you suggested — the module is re-documented as an illustrative private framing over ironfix-fast's stop-bit primitives, not a conforming FAST 1.1 template, and it now calls out the two departures explicitly: bit 0 isn't the TemplateID copy bit, and no-operator optionals would need nullable encoding rather than a pmap bit. Genuine conformance is crate-level work (no template-XML engine, no nullable encodings yet), so I didn't dress it up as a FAST template in a demo.

ironfix-example/src/demo.rs
joaquinbejar: Fixed in 85e62d0 — execution_report() now emits OrderQty(38)/LastShares(32)/LastPx(31) for Fix40|Fix41|Fix42 and LeavesQty(151) for everything except 4.0, so the 4.1 and 4.2 reports carry their full required set instead of just 151. The module-doc table's 4.1/4.2 column is corrected to 'required' too, and there's a test asserting the full set.

ironfix-example/src/demo.rs
joaquinbejar: Fixed in 85e62d0 — new_order_single() now gates TransactTime(60) behind `!matches!(version, Fix40 | Fix41)`, so it's emitted from 4.2 on where it's required and left off the 4.0/4.1 frames where it's out of schema. The exact 4.1 boundary can't be pinned without FIX41.xml (only 44 is vendored), and 4.1 treats 60 as optional, so leaving it off 4.1 is the safe choice.

ironfix-example/src/demo.rs
joaquinbejar: Fixed in 85e62d0 — 1128 is now stamped inside the header (start_app), immediately after MsgType(35) and ahead of SenderCompID(49) where FIXT.1.1 places it, instead of being appended after SendingTime(52) in the body. Only affects the 5.0-family examples. You were right it was an ordering issue, not duplication — 1128 was always emitted exactly once.

ironfix-example/src/demo.rs
joaquinbejar: Leaving this — IncomingOrder::from_new_order checks the four fields it needs to build and echo the ack (11/55/54/38) and Rejects otherwise, so it never emits a malformed ExecutionReport. Version-aware validation of 21/40/60 and conditional-on-OrdType Price(44) is venue-grade and exceeds even the engine, where the dictionary Validator is wired into neither the engine nor the codec. Under-validating well-formed inbound is demo-appropriate.

ironfix-example/examples/common/mod.rs
joaquinbejar: Leaving this — a TestRequest missing TestReqID(112) is malformed input on a path the demo's own client never generates, and the answer (a Heartbeat with no echo) is itself well-formed, since 112 is only conditionally required when echoing a TestRequest. No malformed frame reaches the wire on the happy path. Rejecting a 112-less TestRequest would be reasonable polish but isn't must-fix.

ironfix-example/src/demo.rs
joaquinbejar: Fixed in 85e62d0 — start() now peeks the number (next_sender_seq) for MsgSeqNum(34) and finish() commits it (try_allocate_sender_seq) only after the frame actually encodes, so a failed encode no longer burns a sequence number and leaves a gap. Low practical impact for the demo's tiny messages, but it's the right property for the reusable DemoSession API.

ironfix-example/examples/fast_server.rs
joaquinbejar: Fixed in 85e62d0 — the market-data sequence number now stops the feed on exhaustion (checked_add → break with a warning) instead of wrapping to 1, and the same fix is applied in fast_server_lockfree.rs. Unreachable at u64::MAX ticks in practice, but examples teach habits and this one was teaching the wrapping-seqnum anti-pattern the file's own comment warns against.

ironfix-example/src/lib.rs
joaquinbejar: Acknowledged — these mods have to be public because the example binaries are separate crates that consume them, so narrowing them would break the build. ironfix-example is outside cargo-semver-checks so there's no CI consequence; the only accurate residue is that the PR's 'no library crate changed' line should note ironfix-example's public surface grew. That delta goes in the PR description for the semver record.

ironfix-example/examples/common/mod.rs
joaquinbejar: Fixed the PossDup half in 85e62d0 — that arm now returns Inbound::Duplicate and serve_connection skips business dispatch, so a duplicate isn't reprocessed. Scoped it to just that arm as you said: the gap behaviour is the documented no-store demo simplification (a real acceptor would ResendRequest, which the demo has nothing to resend from), so I left it. The Process/Duplicate outcome split also makes the two paths clearer.

ironfix-example/src/demo.rs
joaquinbejar: Fixed in 85e62d0 — same fix as the first-pass thread: 38/32/31 are emitted through 4.2 alongside 150/151, and the module-doc `OrderQty/LastShares/LastPx` row is corrected to 'required' for the 4.1–4.2 column, since the old table was wrong there too.

ironfix-example/src/fast_market_data.rs
joaquinbejar: Fixed in 85e62d0 — went with the doc caveat: the module now states it's an illustrative framing, not a FAST template, and that FAST 1.1 reserves the first pmap bit for the TemplateID copy while this demo assigns it to the first optional. The self-consistent-but-non-conforming nature is spelled out so nobody copies it as wire FAST; nullable-encoding support would be the crate-level path to real conformance.

ironfix-example/src/lib.rs
joaquinbejar: Same as the other lib.rs thread — the new pub mods are required for the example binaries to compile against them, and ironfix-example is outside the semver workflow so nothing breaks. The additive public-surface change goes in the PR description for the record.

ironfix-example/examples/common/mod.rs
```diff
@@ -73,67 +182,340 @@ pub fn init_logging() {
...
+                let response = session.logon()?;
+                send(&mut connection, response).await?;
+            }
+            MsgType::TestRequest => {
+                let test_req_id = raw.get_field_str(112);
+                let response = session.heartbeat(test_req_id)?;
+                send(&mut connection, response).await?;
+            }
+            MsgType::Heartbeat => {}
+            MsgType::NewOrderSingle => {
```
joaquinbejar, line 498: 🔴 **Critical**: this dispatch is phase-blind, so a `NewOrderSingle` sent as the first in-sequence frame is accepted and acknowledged before any Logon. Track pre-Logon, active, and logout state, reject application traffic until Logon succeeds, and add a pre-Logon order regression test.

ironfix-example/examples/fix44_server_channel.rs
```diff
@@ -1,325 +1,343 @@
...
-                            break;
-                        }
-                    }
-                }
-            }
-            Err(e) => {
-                error!("Read error: {}", e);
-                break;
-            }
+        let Ok(decision) = reply_rx.await else {
```
joaquinbejar, line 176: 🟠 **Warning**: each socket reader waits for the shared processor before polling the connection again, so one slow business decision stalls Heartbeat and TestRequest handling for that session and the single processor serializes this risk across clients. Keep session reads running with an independent bounded response path, or narrow the architecture claim and add a slow-processor test.

ironfix-example/src/fast_market_data.rs
```diff
@@ -0,0 +1,389 @@
...
+//! FAST template. Two differences in particular:
+//!
+//! * Conforming FAST reserves presence-map bit 0 for the template-identifier
+//!   copy operator. This framing does not: it writes the message id as a plain
+//!   mandatory `uint` and gives bit 0 to the first optional field.
+//! * A conforming optional field that has no field operator is carried by FAST
+//!   *nullable* encoding — the value is shifted so one representation is
+//!   reserved to mean "absent" — not by a presence-map bit. This framing signals
+//!   optional presence with presence-map bits because `ironfix-fast` implements
+//!   neither field operators nor nullable encoding yet.
```
joaquinbejar, line 27: 🟠 **Warning**: this says nullable encoding is absent, but this head exposes and tests `FastEncoder::encode_nullable_uint` and `FastDecoder::decode_nullable_uint`. Narrow the stated gap to the missing nullable string or byte forms and template-driven operator integration so the capability documentation matches the code.

joaquinbejar (commented):
Risk remains high. The current head still accepts application traffic before Logon, the channel example stalls session reads behind shared business processing, and its FAST documentation contradicts the nullable codec already shipped. Visible earlier findings also remain on gap handling, malformed TestRequest acceptance, required order validation, and the unacknowledged public API change. Effective verdict REQUEST_CHANGES. Submitted as COMMENT because this PR is self-authored.

PR #46, opened 2026-07-22
fix(engine): bound resend recovery, validate SendingTime, place CompIDs

### LINKED ISSUE #29: ironfix-engine: unsatisfied ResendRequest never times out; SendingTime unvalidated; CompIDs matched by tag not position
## Overview

Three gaps in the `Initiator`'s recovery and validation behaviour, all found by adversarial review of #9 (PR #25) and deliberately left out of its scope. None is a regression; each was verified against a live `Initiator` driven by a hostile stub acceptor.

## Context

- `ironfix-engine/src/initiator.rs` — the reactor and its resend bookkeeping
- `ironfix-engine/src/wire.rs` — `PeerIdentity::validate`, which already documents the positional caveat
- `doc/fix_operations.md` — states each of these as not implemented

### 1. A ResendRequest is never given up on

A peer that never satisfies a resend but keeps sending gapped frames holds the session open indefinitely. Observed: over six seconds at a one-second heartbeat, the client sent a single ResendRequest and then only Heartbeats; `next_target_seq` stayed pinned, nothing was delivered to the application, and `is_timed_out` stayed false. Gapped frames deliberately refresh the heartbeat clock (a gapped message does prove the peer is alive), so nothing ever escalates.

There is no timeout, no retry, and no bound on how long a session may sit unable to make progress.

### 2. SendingTime (52) is never validated — reject reason 10 unimplemented

`doc/fix_operations.md` enumerates reason 10 (SendingTime accuracy problem) and the engine never produces it. This was excluded from #9 on purpose: a clock-skew tolerance is a policy decision that needs its own typed configuration knob and a defensible default, and guessing one would have been worse than declaring the gap.

### 3. CompIDs are matched by tag, not by position

`PeerIdentity::validate` looks fields up by tag across the whole decoded message, so tags 49/56 appearing **anywhere** satisfy the check. A frame that omits them from the standard header but carries them after the application body is accepted and delivered — on both the reactor and the Logon-ack path. Low severity, since a peer able to place the tags in the body can equally place them in the header, but a message missing required header fields is being accepted rather than rejected.

Closing this needs positional information the decoder does not currently retain.

## Tasks

- [ ] Bound how long an unsatisfied ResendRequest may keep a session open; decide between a timeout, a retry limit, or escalation to Logout, and document the choice
- [ ] Add a typed SendingTime tolerance knob with a defensible default and produce Reject reason 10 outside the window
- [ ] Decide whether the decoder should retain enough positional information to enforce standard-header placement, and enforce it for CompIDs if so

## Acceptance criteria

- [ ] A peer that never satisfies a resend causes a bounded, observable outcome rather than an indefinitely stalled session — covered by a stub-acceptor test
- [ ] A message whose SendingTime falls outside the configured window is rejected with reason 10; the window's default and units are documented
- [ ] A frame carrying CompIDs outside the standard header is either rejected, or the acceptance is documented as deliberate with its rationale
- [ ] `doc/fix_operations.md` updated for whichever behaviours become implemented

joaquinbejar:
## Summary

Closes the three recovery gaps adversarial review found against a live `Initiator` — all fixed, none merely documented as a limitation.

Closes #29.

## Stack

- **Base branch:** `stack/18-codegen-derive` (PR #43)

## Gap 1 — an unsatisfied ResendRequest never gave up

A peer that never satisfied a resend but kept sending gapped frames held the session open indefinitely. The reason is subtle and worth stating: gapped frames deliberately refresh the heartbeat clock — a gapped message *does* prove the peer is alive — so the heartbeat timeout can never fire on a stalled gap.

An outstanding request now carries its own clock, independent of the heartbeat. It is retried while attempts remain, then the session is logged out with a `Text (58)` naming the stalled sequence number — an orderly, observable close, not a silent stall or a hard drop. Any in-sequence message clears the state, so the timer measures a gap that is filling *not at all*, never a slow replay.

Defaults: **10 s per attempt, 3 attempts** — an unrecoverable gap is bounded at ~30 s plus the logout handshake. Retry first because a single lost request or a briefly slow store is recoverable; escalate to Logout rather than hard-disconnect because FIX wants an orderly teardown and the peer deserves to know why.

## Gap 2 — SendingTime (52) was never validated

Reject reason 10 was enumerated in the spec and never produced. It is checked now on every inbound frame and on the Logon ack, **before** the heartbeat is refreshed — a wrong clock must not keep a session alive either. Outside the window is reason 10, absent is reason 1, unparseable is reason 6; each is followed by a Logout, because a wrong clock is systemic rather than per-message.

**Default 120 s, symmetric.** This is where FIX engines have converged (QuickFIX `MaxLatency`): an NTP-synced host stays within milliseconds, so 120 s is orders of magnitude more slack than a healthy peer needs, while an unsynced host drifts past it within days. Tighter starts killing sessions over ordinary drift plus queueing; looser stops distinguishing a wrong clock from a right one. A zero window disables the whole check — presence, format and skew together — for a peer with a known clock problem.

The comparison uses a self-contained `UTCTimestamp` parser (no new dependency): it accepts every FIX precision including leap seconds, rejects impossible dates and pre-1970, and round-trips against the formatter across leap days and a century boundary in tests.

## Gap 3 — CompIDs were matched by tag, not position

`PeerIdentity::validate` looked fields up by tag across the whole message, so 49/56 appearing **anywhere** satisfied the check — a frame carrying them only after the application body was accepted. This is **closed, not documented away**: `RawMessage` already preserves wire order, so no change to the decoder was needed. Each identity field must now carry the expected value *and* sit in the standard header; a misplaced one is Reject reason 9 on both the reactor and Logon-ack paths. The header-tag set covers FIX 4.0–5.0 SP2.

## What must not regress

The inbound path from #25/#37 is intact and still tested: GapFill vs Reset, bounded ResendRequest reply with `MsgSeqNum = BeginSeqNo`, inbound `141=Y` requiring MsgSeqNum 1, checked sequence arithmetic, and the rejected-in-sequence-GapFill-consumes-its-number path.

`SessionConfig` gains three fields and four setters; `EngineError` gains `SendingTime`. Workspace is at 0.4.0. `SessionConfig`'s all-public fields without `#[non_exhaustive]` mean adding fields breaks an external exhaustive struct literal — worth a note, though construction is via `new()` plus builders in practice.

## Tests

Workspace at **516 passing**, including a stub-acceptor test per gap: a peer that only holds the socket open getting exactly 3 ResendRequests then a Logout (with a 30 s heartbeat, so the close is attributable to the resend bound alone), a stale SendingTime producing `373=10`/`371=52`, and body-only CompIDs producing `373=9`/`371=49`.

## Verification

```
cargo test --workspace                                    # 516 passed
```

### DIFF
--- a/ironfix-engine/src/initiator.rs
+++ b/ironfix-engine/src/initiator.rs
@@ -20,10 +20,27 @@
 //! # Inbound conformance
 //!
 //! Every inbound frame is checked in this order: decode, `MsgSeqNum` (34)
-//! presence, counterparty identity (49/56, plus 50/57 when configured), then
-//! heartbeat bookkeeping, then sequence validation. Identity comes before the
-//! heartbeat clock deliberately — traffic from the wrong counterparty must not
-//! keep the session alive, reach the [`Application`], or move sequence state.
+//! presence, counterparty identity (49/56, plus 50/57 when configured),
+//! `SendingTime` (52) accuracy, then heartbeat bookkeeping, then sequence
+//! validation. Identity and clock come before the heartbeat bookkeeping
+//! deliberately — traffic from the wrong counterparty, or from one whose clock
+//! is wrong, must not keep the session alive, reach the [`Application`], or
+//! move sequence state.
+//!
+//! Both of those failures end the session after a Reject and a Logout, because
+//! both are systemic rather than per-message: a peer that is cross-wired, or
+//! whose clock is two minutes out, will be so on its next message too.
+//!
+//! # Bounded recovery
+//!
+//! A gap is answered with a `ResendRequest` (2), and a request that goes
+//! unanswered does not wait forever: it is retried every
+//! [`SessionConfig::resend_timeout`] up to
+//! [`SessionConfig::resend_attempt_limit`] attempts, after which the session is
+//! logged out. Without that bound a peer that keeps sending gapped frames holds
+//! a session open indefinitely — the frames refresh the heartbeat clock, so
+//! nothing times out, while the inbound expectation never moves and nothing
+//! reaches the application.
 //!
 //! # Heartbeats
 //!
@@ -119,7 +136,10 @@ use crate::application::{Application, NoOpApplication, RejectReason, SessionId};
 use crate::connection::{Command, Connection, SessionRuntime};
 use crate::error::EngineError;
 use crate::outbound;
-use crate::wire::{self, MessageFactory, PeerIdentity, PendingMessage, UnsupportedVersion};
+use crate::wire::{
+    self, MessageFactory, PeerIdentity, PendingMessage, SendingTimeGuard, SendingTimeProblem,
+    UnsupportedVersion,
+};
 use bytes::{Bytes, BytesMut};
 use futures_util::{SinkExt, StreamExt};
 use ironfix_core::error::{DecodeError, EncodeError};
@@ -136,7 +156,7 @@ use ironfix_store::MessageStore;
 use ironfix_transport::FixCodec;
 use std::num::NonZeroU64;
 use std::sync::{Arc, Mutex};
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use tokio::net::{TcpStream, ToSocketAddrs};
 use tokio::sync::{mpsc, watch};
 use tokio::task::JoinHandle;
@@ -572,6 +592,7 @@ impl<A: Application + 'static> Initiator<A> {
         });
         let mut factory = MessageFactory::new(&self.config, version);
         let identity = PeerIdentity::new(&self.config);
+        let sending_time = SendingTimeGuard::new(&self.config);
         let write_timeout = self.write_timeout;
         let store = self.store.as_ref();
 
@@ -618,7 +639,8 @@ impl<A: Application + 'static> Initiator<A> {
             Ok(Some(Ok(frame))) => frame,
         };
 
-        let mut pending_resend = None;
+        // (gap start, high-water) of a gap detected in the Logon ack itself.
+        let mut pending_resend: Option<(u64, u64)> = None;
         {
             let raw = wire::decode_frame(&ack)?;
             match raw.msg_type() {
@@ -696,6 +718,41 @@ impl<A: Application + 'static> Initiator<A> {
                 .await;
                 let _ = session.on_logon_reject();
                 return Err(EngineError::IdentityMismatch { detail });
+            }
+
+            // The clock is checked on the ack for the same reason it is checked
+            // on every later frame: a peer whose SendingTime is wildly skewed
+            // cannot be trusted to sequence a session, and the handshake is the
+            // cheapest place to refuse it.
+            if let Err(problem) = sending_time.validate(&raw) {
+                let detail = problem.to_string();
+                let reason = problem.reject_reason();
+                let reject = factory.session_reject(ack_seq, MsgType::Logon.as_str(), &reason);
+                let _ = send_handshake_admin(
+                    self.application.as_ref(),
+                    &session_id,
+                    &mut framed,
+                    &mut factory,
+                    &runtime.sequences,
+                    store,
+                    write_timeout,
+                    reject,
+                )
+                .await;
+                let logout = factory.logout(Some(&detail));
+                let _ = send_handshake_admin(
+                    self.application.as_ref(),
+                    &session_id,
+                    &mut framed,
+                    &mut factory,
+                    &runtime.sequences,
+                    store,
+                    write_timeout,
+                    logout,
+                )
+                .await;
+                let _ = session.on_logon_reject();
+                return Err(EngineError::SendingTime { detail });
             }
 
             if let Err(reason) = self.application.from_admin(&raw, &session_id).await {
@@ -867,7 +924,9 @@ impl<A: Application + 'static> Initiator<A> {
                         "logon ack MsgSeqNum too low: expected {expected}, received {received}"
                     )));
                 }
-                SequenceResult::Gap { expected, .. } => pending_resend = Some(expected),
+                SequenceResult::Gap { expected, received } => {
+                    pending_resend = Some((expected, received));
+                }
             }
         }
 
@@ -877,7 +936,7 @@ impl<A: Application + 'static> Initiator<A> {
         tracing::info!(session = %session_id, "FIX session established");
 
         // A gap in the Logon ack means we missed messages: request a resend.
-        if let Some(expected) = pending_resend {
+        if let Some((expected, _high_water)) = pending_resend {
             let request = factory.resend_request(expected, 0);
             send_handshake_admin(
                 self.application.as_ref(),
@@ -907,15 +966,18 @@ impl<A: Application + 'static> Initiator<A> {
             reject_tx,
         ));
 
+        let resend = pending_resend
+            .map(|(expected, high_water)| ResendState::first(expected, high_water, &self.config));
         let reactor = Reactor {
             factory,
             identity,
+            sending_time,
             runtime: Arc::clone(&runtime),
             config: self.config.clone(),
             application: Arc::clone(&self.application),
             session_id: session_id.clone(),
             store: self.store.clone(),
-            pending_resend,
+            resend,
             write_timeout,
             app_tx,
         };
@@ -1209,12 +1271,90 @@ struct ReactorChannels {
     dispatcher: JoinHandle<()>,
 }
 
+/// Bookkeeping for an outstanding `ResendRequest` (2).
+///
+/// A resend that is never satisfied must not keep the session open forever:
+/// gapped frames refresh the heartbeat clock, so the heartbeat timeout never
+/// fires while the inbound expectation sits pinned. This tracks the gap so the
+/// reactor can retry the request and, once the attempts are spent, log the
+/// session out instead of waiting on a peer that is not answering.
+#[derive(Debug, Clone, Copy)]
+struct ResendState {
+    /// The next sequence number the recovery is waiting to receive. It advances
+    /// with in-order inbound progress so a retry re-requests from the current
+    /// gap start, never a number already consumed.
+    expected: u64,
+    /// The top of the outstanding range: the highest sequence number seen while
+    /// the gap has been open. The request is open-ended (`EndSeqNo` = 0), so
+    /// this is the water mark the expectation must cross for the gap to be
+    /// considered closed — clearing recovery on the first in-order message
+    /// instead would forget that later numbers in the range are still missing.
+    high_water: u64,
+    /// When the most recent request for this gap was sent.
+    requested_at: Instant,
+    /// How many requests have been sent for this gap, counting the first.
+    attempts: u32,
+    /// Attempt ceiling, resolved once from the configuration.
+    limit: u32,
+    /// How long a request may make no progress before it is retried.
+    timeout: Duration,
+}
+
+impl ResendState {
+    /// Opens recovery for a gap at `expected`, with `high_water` the highest
+    /// sequence number outstanding, counting the first request.
+    #[must_use]
+    fn first(expected: u64, high_water: u64, config: &SessionConfig) -> Self {
+        Self {
+            expected,
+            high_water,
+            requested_at: Instant::now(),
+            attempts: 1,
+            limit: config.resend_attempt_limit(),
+            timeout: config.resend_timeout,
+        }
+    }
+
+    /// Whether this gap has waited longer than its retry timeout.
+    #[must_use]
+    fn is_stalled(&self) -> bool {
+        self.requested_at.elapsed() >= self.timeout
+    }
+
+    /// Whether another request may still be sent for this gap.
+    #[must_use]
+    const fn can_retry(&self) -> bool {
+        self.attempts < self.limit
+    }
+
+    /// Records that another request for this same gap has gone out.
+    fn record_retry(&mut self) {
+        self.requested_at = Instant::now();
+        self.attempts = self.attempts.saturating_add(1);
+    }
+
+    /// Records genuine in-order progress within the outstanding range.
+    ///
+    /// The gap start has advanced to `expected`, so the stall clock and the
+    /// retry budget both reset: a peer that is actively feeding the gap has
+    /// proved itself responsive and must not be logged out for being slow. A
+    /// duplicate or an out-of-order frame makes no such progress and so does not
+    /// reach here.
+    fn record_progress(&mut self, expected: u64) {
+        self.expected = expected;
+        self.requested_at = Instant::now();
+        self.attempts = 1;
+    }
+}
+
 /// Reactor state shared across event handlers.
 struct Reactor<A: Application> {
     /// Outbound frame factory, holding the encoder reused for every message.
     factory: MessageFactory,
     /// Identity every inbound message must carry.
     identity: PeerIdentity,
+    /// Clock-skew check applied to every inbound `SendingTime` (52).
+    sending_time: SendingTimeGuard,
     /// Shared session runtime (sequences + heartbeat).
     runtime: Arc<SessionRuntime>,
     /// Session configuration.
@@ -1225,12 +1365,12 @@ struct Reactor<A: Application> {
     session_id: SessionId,
     /// Optional message store for outbound frames and resend replay.
     store: Option<Arc<dyn MessageStore>>,
-    /// Expected sequence number a pending ResendRequest was issued for.
+    /// The outstanding ResendRequest, if a gap is being recovered.
     ///
     /// The logout deadline is *not* tracked here: `Session<LogoutPending>`
     /// carries the instant the Logout went out, so the phase itself is the
     /// deadline.
-    pending_resend: Option<u64>,
+    resend: Option<ResendState>,
     /// Bound on a single socket write.
     write_timeout: Duration,
     /// Hand-off queue to the application dispatcher.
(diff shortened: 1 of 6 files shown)


joaquinbejar (commented):
Effective verdict: REQUEST_CHANGES. The recovery timer can still be bypassed after partial progress, and the new SendingTime validation accepts protocol-invalid headers. Submitted as COMMENT because this is a self-authored pull request. CI is green and the crate DAG is intact; the new public SessionConfig fields are source-breaking but fit the planned 0.4.0 release.

ironfix-engine/src/initiator.rs
```diff
@@ -603,155 +1308,149 @@ impl<A: Application> Reactor<A> {
...
         }
 
         match self.runtime.sequences.validate_incoming(seq) {
             SequenceResult::Ok => {
-                self.runtime.sequences.increment_target_seq();
-                self.pending_resend = None;
+                if let Err(err) = self.runtime.sequences.try_increment_target_seq() {
+                    return Err(exhausted(phase, err));
+                }
+                self.resend = None;
```
joaquinbejar: 🔴 Critical: one replayed message clears all knowledge of the still-open gap. For example, after seq 5 opens a gap at 2, receiving seq 2 clears this state; repeated timely 43=Y duplicates of seq 2 then refresh the heartbeat before the TooLow branch returns, so neither heartbeat nor resend recovery closes the stalled session. Track the original gap high-water mark and update/restart recovery on progress until the entire range is closed; add this partial-progress regression and correct the policy stated in doc/fix_operations.md.

ironfix-engine/src/wire.rs
```diff
@@ -14,30 +14,486 @@
...
+    /// whole, so an operator who has switched it off for a peer with a known
+    /// clock problem does not get half of it back.
+    ///
+    /// # Errors
+    /// Returns [`SendingTimeProblem::Missing`] when tag 52 is absent,
+    /// [`SendingTimeProblem::Malformed`] when it is not a `UTCTimestamp`, and
+    /// [`SendingTimeProblem::Skewed`] when it parses but lies further from the
+    /// local clock than the configured tolerance, in either direction.
+    pub(crate) fn validate(&self, raw: &RawMessage<'_>) -> Result<(), SendingTimeProblem> {
+        if self.is_disabled() {
```
joaquinbejar, line 396: 🔴 Critical: zero tolerance bypasses required-field and format validation, so a peer may omit standard-header SendingTime(52) or send garbage and still establish and retain the session. Tag 52 remains required regardless of latency policy. Let zero disable only the skew comparison, while presence, position, and UTCTimestamp parsing remain enforced, and correct the contrary policy in doc/fix_operations.md.

ironfix-engine/src/wire.rs
```diff
@@ -14,30 +14,486 @@
...
+            return Ok(());
+        }
+        let Some(value) = raw.get_field_str(52) else {
+            return Err(SendingTimeProblem::Missing);
```
joaquinbejar, line 402: 🔴 Critical: get_field_str searches the entire decoded message, so a current tag 52 placed only after the body begins satisfies this guard even though the required standard-header SendingTime is missing. Require tag 52 within the leading standard-header run before parsing it, and add both Logon-ack and active-session body-only tests.

ironfix-engine/src/wire.rs
```diff
@@ -14,30 +14,486 @@
...
+}
+
+impl SendingTimeGuard {
+    /// Builds the guard from the session configuration.
+    #[must_use]
+    pub(crate) fn new(config: &SessionConfig) -> Self {
+        // `as_millis` is u128 because a Duration can outrun a u64 of
+        // milliseconds; a tolerance that large is indistinguishable from the
+        // largest one this can hold, so it is clamped rather than wrapped.
+        let tolerance_millis =
```
joaquinbejar: 🟠 Warning: as_millis truncates every nonzero tolerance below 1 ms to zero, and is_disabled then silently turns validation off. That contradicts the documented Duration::ZERO-only switch. Preserve Duration precision, or explicitly round a nonzero value up without mapping it to disabled; add a nanosecond-tolerance test.

ironfix-engine/src/reactor.rs
```diff
@@ -0,0 +1,1044 @@
...
+   Author: Joaquín Béjar García
+   Email: jb@taunais.com
+   Date: 22/7/26
+******************************************************************************/
+
+//! The live-session reactor: the inbound/outbound event loop shared by the
+//! [`Initiator`](crate::Initiator) and the [`Acceptor`](crate::Acceptor).
+//!
+//! Both engines establish a session with a role-specific Logon handshake and
+//! then hand the framed socket to exactly this reactor. Everything that keeps a
```
joaquinbejar: 🟠 Warning: this 1,044-line reactor is not declared by lib.rs, while initiator.rs still contains and uses its own Reactor and no Acceptor exists. The file is therefore never compiled, linted, or tested, and its claim that both engines use it is false. Remove the duplicate or complete and wire the extraction without claiming the missing Acceptor.

ironfix-engine/src/wire.rs
joaquinbejar, line 396: 🔴 **Critical**: zero tolerance returns success before checking that required `SendingTime(52)` exists, parses as `UTCTimestamp`, and sits in the standard header; the later whole-message lookup also accepts a body-only tag. Zero should disable only skew comparison. Always require and parse a header-positioned 52 so missing and malformed values still produce reasons 1 and 6.

ironfix-engine/src/reactor.rs
```diff
@@ -0,0 +1,1044 @@
+/******************************************************************************
+   Author: Joaquín Béjar García
+   Email: jb@taunais.com
+   Date: 22/7/26
+******************************************************************************/
+
+//! The live-session reactor: the inbound/outbound event loop shared by the
```
joaquinbejar: 🟠 **Warning**: this new 1,044-line module is never declared by `lib.rs` or used by `initiator.rs`, which still owns its own reactor, so none of this code is compiled, linted, or tested. It also claims a shared `Acceptor` that does not exist. Remove the file or complete the refactor without documenting the missing Acceptor as built.

ironfix-engine/src/initiator.rs
```diff
@@ -391,10 +571,76 @@ fn teardown(phase: Phase) {
...
+    /// Whether another request may still be sent for this gap.
+    #[must_use]
+    const fn can_retry(&self) -> bool {
+        self.attempts < self.limit
+    }
+
+    /// Records that another request for this same gap has gone out.
+    fn record_retry(&mut self) {
+        self.requested_at = Instant::now();
+        self.attempts = self.attempts.saturating_add(1);
```
joaquinbejar, line 1333: 🟠 **Warning**: `saturating_add` is prohibited here because it hides retry-accounting overflow and future invariant drift. The current guard makes overflow unreachable, so use `checked_add` and surface an explicit terminal outcome; fix the duplicate in the dead reactor module too.

joaquinbejar (commented):
Risk is high. The recovery logic is bounded, but SendingTime validation can waive a required header field and the branch also ships an uncompiled duplicate reactor with hidden arithmetic. Effective verdict REQUEST_CHANGES. Submitted as COMMENT because this PR is self-authored.

ironfix-engine/src/wire.rs
```diff
@@ -14,30 +14,486 @@
...
+    /// # Errors
+    /// Returns [`SendingTimeProblem::Missing`] when tag 52 is absent,
+    /// [`SendingTimeProblem::Malformed`] when it is not a `UTCTimestamp`, and
+    /// [`SendingTimeProblem::Skewed`] when it parses but lies further from the
+    /// local clock than the configured tolerance, in either direction.
+    pub(crate) fn validate(&self, raw: &RawMessage<'_>) -> Result<(), SendingTimeProblem> {
+        if self.is_disabled() {
+            return Ok(());
+        }
+        let Some(value) = raw.get_field_str(52) else {
```
joaquinbejar: 🔴 **Critical**: `get_field_str(52)` searches the whole message, so a current timestamp placed after the first body field passes even though the required standard-header `SendingTime` is absent. Use a header-scoped lookup and add body-only tag 52 tests on both Logon and active-session paths.

ironfix-engine/src/initiator.rs
```diff
@@ -574,8 +887,403 @@ impl<A: Application> Reactor<A> {
...
+                Phase::Active(session) => self.begin_logout(framed, session, Some(&reason)).await,
+            };
+        }
+
+        let expected = resend.expected;
+        let out_seq = match self.runtime.sequences.try_allocate_sender_seq() {
+            Ok(seq) => seq.value(),
+            Err(err) => return Err(exhausted(phase, err)),
+        };
+        let frame = self.factory.resend_request(out_seq, expected, 0);
```
joaquinbejar: 🔴 **Critical**: rejected or invalid in-sequence GapFills consume their sequence number but leave this stored `expected` unchanged. The timer then requests an already-consumed sequence and can log out despite recovery progress. Synchronize or clear `ResendState` whenever those paths consume a GapFill and add that recovery regression.

ironfix-engine/src/wire.rs
```diff
@@ -14,30 +14,486 @@
...
+
+impl SendingTimeGuard {
+    /// Builds the guard from the session configuration.
+    #[must_use]
+    pub(crate) fn new(config: &SessionConfig) -> Self {
+        // `as_millis` is u128 because a Duration can outrun a u64 of
+        // milliseconds; a tolerance that large is indistinguishable from the
+        // largest one this can hold, so it is clamped rather than wrapped.
+        let tolerance_millis =
+            u64::try_from(config.sending_time_tolerance.as_millis()).unwrap_or(u64::MAX);
```
joaquinbejar: 🟠 **Warning**: every positive tolerance below one millisecond truncates to zero here, and `is_disabled` then turns validation off entirely. Preserve explicit `Duration::ZERO` as the only disable state and retain or round up positive sub-millisecond values, with a regression test.

ironfix-engine/src/initiator.rs
```diff
@@ -574,8 +887,403 @@ impl<A: Application> Reactor<A> {
...
+            let frame =
+                self.factory
+                    .session_reject(out_seq.value(), ref_seq, ref_msg_type, &reason);
+            let _ = self.send_admin(framed, frame).await;
+        }
+        if let Ok(out_seq) = self.runtime.sequences.try_allocate_sender_seq() {
+            let frame = self.factory.logout(out_seq.value(), Some(&detail));
+            let _ = self.send_admin(framed, frame).await;
+        }
+        teardown(phase);
```
joaquinbejar, line 2072: 🟠 **Warning**: this sends Logout and immediately tears the session down, so it never enters `LogoutPending` or waits for the peer acknowledgment through `logout_timeout`. Route this failure through the normal Logout handshake and test both acknowledgment and timeout.

ironfix-engine/src/reactor.rs
joaquinbejar: 🟠 **Warning**: `lib.rs` never declares this module, so none of these 1,044 lines is compiled or tested while `initiator.rs` still contains the active duplicate reactor. Complete the extraction atomically or remove this dead file; the claim that both engines share it is false on this head.

joaquinbejar (commented):
Risk is high. Effective verdict REQUEST_CHANGES. CI is green, but gap recovery can retry a consumed sequence, body-only SendingTime is accepted, sub-millisecond tolerances disable validation, the SendingTime Logout bypasses the handshake, and the new shared reactor is dead code. The branch also predates the exact-head fix from its declared PR 43 base. Submitted as COMMENT because this PR is self-authored.

ironfix-engine/src/wire.rs
```diff
@@ -14,30 +14,486 @@
...
+            | 145
+            | 212
+            | 213
+            | 347
+            | 369
+            | 627
+            | 628
+            | 629
+            | 630
+            | 1128
```
joaquinbejar, line 180: 🟠 **Warning**: this version-blind union lets a later-version header tag extend the apparent header on an older session. A FIX.4.4 frame with `1128` before `49` and `56` therefore passes the positional CompID check even though `1128` is FIXT-only. Resolve header membership from the negotiated FIX version and add a cross-version regression fixture.

ironfix-engine/src/wire.rs
joaquinbejar: 🟠 **Warning**: `get_field_str` conflates an absent tag with a present non-UTF-8 value. The decoder accepts arbitrary field bytes, so `52=\xFF` emits reason 1 even though the PR documents every present, unparseable SendingTime as reason 6. Inspect the raw field first, map UTF-8 failure to `Malformed`, and add a raw-byte regression asserting `371=52` and `373=6`.

ironfix-engine/src/wire.rs
```diff
@@ -14,30 +14,486 @@
...
+    for byte in bytes {
+        value = value.checked_mul(10)?.checked_add(u64::from(byte - b'0'))?;
+    }
+    Some(value)
+}
+
+/// Parses the fractional-second digits of a `UTCTimestamp` into milliseconds,
+/// truncating anything below a millisecond.
+fn parse_fraction_millis(bytes: &[u8]) -> Option<u64> {
+    if bytes.is_empty() || bytes.len() > 9 || !bytes.iter().all(u8::is_ascii_digit) {
```
joaquinbejar, line 505: 🟠 **Warning**: this nine-digit ceiling rejects the 12-digit picosecond `UTCTimestamp` defined by the FIX Time Precision Addendum for supported FIX variants, so a conforming peer can receive reason 6 and Logout. Accept 12 fractional digits while truncating only for the millisecond comparison, and cover the form in both Logon and active-session tests.

joaquinbejar (commented):
Risk remains high. The dead reactor finding is fixed and focused engine and session tests pass, but the current head leaves the existing blocking threads unresolved and adds three protocol compatibility defects. Effective verdict REQUEST_CHANGES. Submitted as COMMENT because this PR is self-authored.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed in 9570e29 — this was the top finding. ResendState now carries a high-water mark, and note_inbound_progress clears recovery only once the expectation crosses the whole outstanding range, advancing the gap start (and resetting the stall clock and retry budget) on genuine in-order progress instead of wiping the state on the first message back. A duplicate takes the dedup path and never touches it, so the partial-fill-then-duplicate-flood now escalates to Logout — there's a test that fails against the old behaviour.

ironfix-engine/src/wire.rs
joaquinbejar: This one's spec-endorsed — doc/fix_operations.md:335-336 explicitly says setting the tolerance to Duration::ZERO disables the check including its presence and format checks, so returning Ok when is_disabled() conforms. The default is 120s, so presence/format are checked unless an operator deliberately opts out. The body-only-52 half is real and is handled on the separate positional thread.

ironfix-engine/src/wire.rs
joaquinbejar: Fixed in 9570e29 — SendingTimeGuard::validate now looks tag 52 up header-scoped (get_field(52) filtered through in_standard_header), the same positional guard #46 added for the CompIDs, so a body-only 52 reads as missing and a header-omitted one isn't silently satisfied. Severity's the P2/P3 you noted — the binding spec assigns no positional reject reason for 52 — so it's inbound-acceptance hardening rather than a wire-emit defect.

ironfix-engine/src/wire.rs
joaquinbejar: Fixed in 9570e29 — the guard stores the Duration now instead of flattening to whole milliseconds, and is_disabled() is Duration::is_zero(), so a sub-millisecond tolerance like 500us stays enabled instead of truncating to 0 and colliding with the disable sentinel. Skew is compared as a Duration too, and there's a test on the 500us case.

ironfix-engine/src/reactor.rs
joaquinbejar: reactor.rs doesn't exist on this branch — the engine module tree is application/builder/connection/error/initiator/lib/outbound/wire. This looks like it was flagged against a later branch in the stack; nothing to change here.

ironfix-engine/src/wire.rs
joaquinbejar: Same as the other zero-tolerance thread — doc/fix_operations.md:335-336 endorses Duration::ZERO skipping presence and format, so the code conforms to the binding spec. Stands.

ironfix-engine/src/reactor.rs
joaquinbejar: Same as the other reactor.rs thread — that module doesn't exist on this branch, so there's nothing to change here.

ironfix-engine/src/initiator.rs
joaquinbejar: This stands — `attempts` is a bounded u32 retry counter, not a sequence number. The CLAUDE.md 🔴 ban is specifically saturating/wrapping on a seqnum because it corrupts a trading session; a retry counter that's already bounded below its limit isn't that, and as you noted the overflow is unreachable anyway.

ironfix-engine/src/wire.rs
joaquinbejar: Fixed in 9570e29 — same header-scoping fix: tag 52 is now looked up via get_field(52) filtered through in_standard_header, so a body-only 52 no longer passes and a header-omitted one is caught. Same P2/P3 severity, since the binding spec mandates no positional reason for 52.

ironfix-engine/src/initiator.rs
joaquinbejar: Fixed in 9570e29 — same root as the top finding. The three in-sequence GapFill reject paths now go through consume_rejected_fill, which reconciles the outstanding resend whenever the fill actually advanced the target, so a fill that errors mid-recovery advances the gap start instead of leaving on_resend_stalled to re-ask for an already-consumed number and eventually log the session out.

ironfix-engine/src/wire.rs
joaquinbejar: Fixed in 9570e29 — same truncation fix: the guard stores the Duration and is_disabled() is is_zero(), so a positive sub-millisecond tolerance stays on instead of flooring to 0 and disabling the whole check.

ironfix-engine/src/initiator.rs
joaquinbejar: This stands — a wrong-clock or cross-wired peer is a systemic error, and doc/fix_operations.md:98-100 and 123-125 prescribe exactly Reject + Logout + drop for those, not the graceful LogoutPending handshake. Routing it through the logout_timeout path would actually be wrong here.

ironfix-engine/src/reactor.rs
joaquinbejar: Same as the other reactor.rs threads — the module doesn't exist on this branch, so there's nothing to change.

ironfix-engine/src/wire.rs
joaquinbejar: This is deliberate — is_standard_header_tag unions the FIX 4.0–5.0 SP2 header set by design (it's stated in the fn doc). A FIXT-only tag showing up in a 4.4 header just means the header run isn't broken at that tag; the CompID positional check still passes for a well-formed header, and the binding spec is silent on cross-version header membership. Version-resolved membership would be a future enhancement, not a current defect.

ironfix-engine/src/wire.rs
joaquinbejar: Fixed in 9570e29 — validate() now pulls the raw field bytes and maps a present-but-non-UTF-8 52 to Malformed (reason 6) rather than folding it into Missing (reason 1), matching doc/fix_operations.md:320 for a present-but-unparseable SendingTime. A test asserts the reason code is 6.

ironfix-engine/src/wire.rs
joaquinbejar: Leaving this — 9 fractional digits (nanosecond) covers every precision FIX 4.0–5.0 SP2 puts on the wire in practice, the binding spec sets no fractional-digit limit, and the 12-digit picosecond form is a rarely-emitted optional addendum. It's a defensible ceiling; worth revisiting only if we certify against a picosecond-precision venue.

PR #47, opened 2026-07-22
feat(engine): implement the Acceptor and give EngineBuilder a real terminal

### LINKED ISSUE #21: ironfix-engine: no Acceptor and an inert EngineBuilder — advertised server-side engine, TLS, and builder entry point do not exist; roadmap decisions needed
## Overview

ironfix-engine's published docs advertise an Acceptor that does not exist — the session crate's fully modeled acceptor FSM has zero consumers and all 9 server examples hand-roll accept loops with no real session behavior — and EngineBuilder is a published API whose every setter is a no-op (no build()/terminal method, TLS/reconnect/timeout knobs consumed by nothing), the root cause of the broken README quick start. These need explicit decisions: implement (Acceptor mirroring Initiator; build() as a supervisor) or shrink the surface and fix the docs.

## Context

Crates / areas touched: `ironfix-engine`, `ironfix-session`, `ironfix-transport`

### 1. Known gap: no Acceptor — docs claim one, acceptor FSM has zero consumers (finding 55)
`ironfix-engine/src/lib.rs:13`, `ironfix-example/examples/fix40_server.rs:45-55`

lib.rs: '- **Acceptor**: Server-side FIX engine for accepting connections' — only Initiator exists. ironfix-session fully models accept()/on_logon_received()/accept_logon()/reject_logon()/on_timeout() with no caller anywhere (grep over engine/example: zero call sites outside state.rs tests). Server examples duplicate raw accept loops with no inbound seqnum validation, no CompID checks, no heartbeat initiation (a real HeartBtInt=30 counterparty would time them out), no Logout handshake.

**Fix:** short-term, fix the crate docs. Roadmap: Acceptor<A> mirroring Initiator (per-connection Reactor over Framed<TcpStream, FixCodec>, driving the acceptor typestate, logon_timeout via LogonReceived.received_at, CompID auth), port one fixNN_server onto it, update the canon row.

### 2. Known gap: EngineBuilder is inert — no terminal method, all knobs consumed by nothing (finding 74)
`ironfix-engine/src/builder.rs:18`

EngineBuilder collects application, sessions, use_tls, connect_timeout, reconnect knobs — nothing in the workspace reads them (pub fns: new/with_application/add_session/with_connect_timeout/with_reconnect_interval/sessions/application — no terminal method; no consumer outside its own tests). Initiator::new takes SessionConfig + Arc<A> directly and its docs declare reconnection out of scope, contradicting the builder's knobs; with_tls silently does nothing (no TLS exists). A published API whose every setter is a no-op, and the root cause of the broken README quick start.

**Fix:** either implement build() as a supervisor honoring the knobs (TLS gated on transport rustls), or shrink the builder to what is real and fix the README; until decided, rustdoc must state it is configuration-only.

## Tasks

- [ ] Known gap: no Acceptor — docs claim one, acceptor FSM has zero consumers (finding 55)
- [ ] Known gap: EngineBuilder is inert — no terminal method, all knobs consumed by nothing (finding 74)

## Acceptance criteria

- [ ] An explicit decision is recorded (ADR or issue resolution) for each: implement Acceptor vs. remove the doc claim; implement EngineBuilder::build() as a supervisor vs. shrink the builder
- [ ] Until implementation lands, ironfix-engine crate docs no longer advertise an Acceptor and EngineBuilder rustdoc states it is configuration-only with no runtime effect
- [ ] If Acceptor is implemented: it drives the ironfix-session acceptor typestate (logon timeout, CompID auth), at least one fixNN_server example is ported onto it, and an e2e test drives Initiator against Acceptor
- [ ] If build() is implemented: every existing knob (TLS, connect_timeout, reconnect) is honored or removed; with_tls either works via a transport TLS feature or is deleted
- [ ] The 'Design canon vs. as-built' rows for Acceptor and EngineBuilder are updated to match the outcome

joaquinbejar:
## Summary

Implements the Acceptor the crate advertised but never had, and gives `EngineBuilder` a terminal method that produces a runnable engine. Roadmap decisions taken with the maintainer: **implement** the Acceptor, **shrink** the builder to what is real.

Closes #21.

## Stack

- **Base branch:** `stack/29-engine-recovery` (PR #46)
- The last of the stack. Merge bottom-up.

## The Acceptor

`ironfix-session` fully modelled the acceptor typestate with **zero consumers**, and the nine server examples hand-rolled accept loops with no sequence validation, no CompID check and no heartbeat — a real `HeartBtInt=30` counterparty would time them out.

`Acceptor<A>` mirrors `Initiator<A>`. Crucially, **the inbound path is not forked**: the reactor machinery is extracted into a shared `reactor` module that both `Initiator` and `Acceptor` hand their framed socket to, so sequence validation, GapFill/Reset, CompID and SendingTime checks, heartbeat, ResendRequest and Logout are single-sourced. A divergent copy would have rotted. Only the acceptor-half handshake is new: await the inbound Logon within `logon_timeout`, drive `accept → on_logon_received → accept_logon`/`reject_logon`, validate identity and SendingTime, adopt and echo the counterparty's `HeartBtInt`, mirror `ResetSeqNumFlag`, then go Active. Every sequence increment is checked; credentials (554/925/RawData) are never logged.

One `Acceptor` serves one configured counterparty, symmetric with the single-session `Initiator`. `fix44_server.rs` is ported onto it as a worked reference; the other eight servers are left for a follow-up.

## EngineBuilder

Every setter was a no-op. The knobs nothing could honour are **removed** — `with_tls` above all, since there is no TLS in the workspace, plus the reconnect knobs the `Initiator` docs already declare out of scope. It gains `into_initiator` and `into_acceptor`, which produce a runnable engine, so the README quick start **compiles for the first time**.

## Public surface (semver)

Added (minor): `Acceptor` and its methods, `EngineBuilder::into_initiator`/`into_acceptor`, `EngineError::Configuration`. Removed (breaking): six `EngineBuilder` no-op methods. The workspace is at 0.4.0, which against the published 0.3.1 baseline is already a major bump, so `cargo-semver-checks` covers the removals — worth an `architect`/`semver.yml` confirm regardless.

## e2e proof

`test_initiator_against_acceptor_full_session_completes` drives a real `Initiator` against a real `Acceptor` over loopback — Logon both ways, an application message each way, graceful Logout — which is the proof that the two FSMs agree rather than merely both compiling. Plus tests for CompID rejection, logon timeout, a non-Logon first frame, and ResetSeqNumFlag mirroring.

## Tests

Workspace at **522 passing**. `CLAUDE.md`'s "Design canon vs. as-built" rows for Acceptor and EngineBuilder are updated to reflect reality.

## Verification

```
cargo test --workspace                                    # 522 passed
cargo build --release --workspace --examples              # clean
```


ironfix-engine/src/acceptor.rs
```diff
@@ -0,0 +1,504 @@
...
+        // Typestate: Disconnected -> Connecting (acceptor side).
+        let session = Session::<Disconnected>::new(session_id.to_string()).accept();
+
+        let _ = stream.set_nodelay(true);
+        let codec = FixCodec::new()
+            .with_max_message_size(self.config.max_message_size)
+            .with_checksum_validation(self.config.validate_checksum);
+        let mut framed = Framed::new(stream, codec);
+
+        let sequences = match self.initial_sequences {
```
joaquinbejar, line 326: 🔴 **Critical**: every `serve` call creates fresh sequence state from the same seed, while the documented reusable `Acceptor` can serve sequential or concurrent sockets for one `SessionId`. Reconnects therefore reuse outbound numbers and forget the inbound expectation, and overlapping sockets can run independent streams for one FIX session. Persist one sequence state per configured session, reject a second active connection, and add reconnect and duplicate-connection tests.

ironfix-engine/src/acceptor.rs
```diff
@@ -0,0 +1,504 @@
...
+        {
+            let raw = match wire::decode_frame(&logon_frame) {
+                Ok(raw) => raw,
+                Err(err) => {
+                    let _ = session.reject_logon();
+                    return Err(err.into());
+                }
+            };
+
+            match raw.msg_type() {
```
joaquinbejar, line 381: 🔴 **Critical**: the handshake accepts Logon without comparing `BeginString(8)` to the configured wire version, validating required `EncryptMethod(98)`, or requiring and matching `DefaultApplVerID(1137)` for FIXT.1.1. A cross-version peer can become Active and receive an acknowledgement in a different protocol version. Validate these fields before callbacks or sequence movement and add FIX 4.x and FIXT mismatch fixtures.

ironfix-engine/src/acceptor.rs
```diff
@@ -0,0 +1,504 @@
...
+                MsgType::Logon => {}
+                other => {
+                    let msg_type = other.as_str().to_string();
+                    let _ = session.reject_logon();
+                    return Err(EngineError::UnexpectedMessage { msg_type });
+                }
+            }
+
+            let logon_seq: u64 = match raw.get_field_as(34) {
```
joaquinbejar: 🔴 **Critical**: `get_field_as(34)` searches the whole message, so a `MsgSeqNum` placed only in the body is accepted as the required standard-header field and can advance session state. Use a header-scoped lookup, as the CompID validation does, and add a body-only tag 34 regression.

ironfix-engine/src/acceptor.rs
```diff
@@ -0,0 +1,504 @@
...
+                }
+                let _ = session.reject_logon();
+                return Err(EngineError::LogonRejected {
+                    reason: reason.text,
+                });
+            }
+
+            // Honor the heartbeat interval the initiator requested (HeartBtInt,
+            // 108): the acceptor adopts it and echoes it back on the reply.
+            let requested_heartbeat_secs =
```
joaquinbejar: 🔴 **Critical**: missing or nonnumeric required `HeartBtInt(108)` silently falls back to local configuration, while `108=18446744073709551615` is accepted as a `Duration`; the first heartbeat check then evaluates `interval + 1s` and panics, aborting the release process. Require and bound this untrusted value, use checked deadline arithmetic, and cover missing, malformed, maximum, and legal zero cases.

ironfix-engine/src/acceptor.rs
```diff
@@ -0,0 +1,504 @@
...
+        let codec = FixCodec::new()
+            .with_max_message_size(self.config.max_message_size)
+            .with_checksum_validation(self.config.validate_checksum);
+        let mut framed = Framed::new(stream, codec);
+
+        let sequences = match self.initial_sequences {
+            Some((sender, target)) if !self.config.reset_on_logon => {
+                SequenceManager::with_initial(sender, target)
+            }
+            _ => SequenceManager::new(),
```
joaquinbejar, line 335: 🔴 **Critical**: local `reset_on_logon` discards configured continuity seeds before the peer's tag 141 is inspected, but the reply only mirrors the peer's flag. If the peer omits `141=Y`, this silently resets both counters to 1 while replying without declaring a reset. Seed first and reset only after a valid negotiated flag, or explicitly signal and enforce the acceptor-initiated reset.

ironfix-engine/src/acceptor.rs
```diff
@@ -0,0 +1,504 @@
...
+                }
+            };
+            let reply = match factory.logon(reply_seq, reply_heartbeat_secs, reset) {
+                Ok(frame) => frame,
+                Err(err) => {
+                    let _ = session.reject_logon();
+                    return Err(err.into());
+                }
+            };
+            if let Ok(mut owned) = wire::owned_from_frame(&reply) {
```
joaquinbejar: 🟠 **Warning**: `to_admin` mutates a temporary `OwnedMessage`, but the untouched original `reply` buffer is sent immediately afterward, so every callback change is discarded. Re-encode and send the callback result, or change the callback contract so it is not falsely mutable, and add a Logon mutation test.

ironfix-example/examples/fix44_server.rs
```diff
@@ -1,186 +1,166 @@
...
+        CompId::new(&cfg.sender_comp_id)?,
+        CompId::new(&cfg.target_comp_id)?,
+        FIX_VERSION,
+    )
+    .with_heartbeat_interval(Duration::from_secs(cfg.heartbeat_interval));
+
+    let app = Arc::new(ServerApp::default());
+    let acceptor = Acceptor::new(config, Arc::clone(&app));
+
+    let connection = acceptor.serve(stream).await?;
```
joaquinbejar, line 199: 🟠 **Warning**: `serve` has already spawned the reactor before line 161 installs the outbound handle. A client can pipeline `NewOrderSingle` immediately after Logon, making `from_app` observe `None` and silently omit the ExecutionReport. Initialize the reply path before inbound dispatch starts, or gate reactor startup, and add a pipelined-order regression.

ironfix-example/examples/fix44_server.rs
```diff
@@ -1,186 +1,166 @@
...
+                .push_str(11, clid)
+                .push_str(17, &format!("EX{clid}"))
+                .push_char(150, '0')
+                .push_char(39, '0')
+                .push_str(55, symbol)
+                .push_str(54, side)
+                .push_str(151, qty)
+                .push_str(14, "0")
+                .push_str(6, "0");
+            if let Err(err) = connection.send(exec).await {
```
joaquinbejar, line 154: 🟠 **Warning**: `from_app` runs on the reactor task but awaits `Connection::send`, whose bounded queue is consumed by that same task. Once the queue fills, the callback waits for capacity that cannot be freed, deadlocking message processing, heartbeat, and Logout. Provide a reactor-local reply path or drain replies on an independent task while preserving ordering and errors.

joaquinbejar (commented):
Risk is high. The new acceptor can lose sequence continuity, accept malformed or cross-version Logons, and panic on hostile heartbeat negotiation, while the worked server can drop or deadlock replies. Effective verdict REQUEST_CHANGES. Submitted as COMMENT because this pull request is self-authored.

ironfix-engine/src/acceptor.rs
```diff
@@ -0,0 +1,504 @@
...
+                let _ = session.reject_logon();
+                return Err(EngineError::LogonRejected {
+                    reason: reason.text,
+                });
+            }
+
+            // Honor the heartbeat interval the initiator requested (HeartBtInt,
+            // 108): the acceptor adopts it and echoes it back on the reply.
+            let requested_heartbeat_secs =
+                raw.get_field_str(108).and_then(|s| s.parse::<u64>().ok());
```
joaquinbejar: 🔴 **Critical**: `108=u64::MAX` is accepted here and installed as a `Duration`. On the next reactor tick, `HeartbeatManager::should_send_test_request` evaluates `interval + 1s`, which overflows and panics; with `panic = "abort"`, a peer can kill the process during Logon. Bound tag 108 against all downstream deadline arithmetic, reject out-of-range values, and add the maximum-value regression.

ironfix-engine/src/acceptor.rs
```diff
@@ -0,0 +1,504 @@
...
+    /// Returns an [`EngineError`] if framing or the Logon handshake fails. That
+    /// includes [`EngineError::LogonTimeout`] when no Logon arrives within
+    /// [`SessionConfig::logon_timeout`], [`EngineError::UnexpectedMessage`] when
+    /// the first frame is not a Logon, [`EngineError::IdentityMismatch`] when
+    /// the Logon's CompIDs do not match the configured counterparty,
+    /// [`EngineError::LogonRejected`] when the `from_admin` authentication hook
+    /// rejects it, and [`EngineError::SequenceExhausted`] when a sequence
+    /// counter has reached `u64::MAX`. A `BeginString` this engine cannot frame
+    /// conformantly is refused up front with [`EngineError::UnsupportedVersion`].
+    pub async fn serve(&self, stream: TcpStream) -> Result<Connection, EngineError> {
```
joaquinbejar, line 300: 🔴 **Critical**: each `serve` call creates independent sequence state, while this Acceptor represents one configured `SessionId` and its docs encourage concurrent calls through `Arc`. Two overlapping valid Logons can therefore both become Active at sequence 1 and deliver duplicate orders through callbacks that cannot distinguish the streams. Add shared per-session admission state that rejects or replaces a duplicate until the first reactor closes, with an overlapping-Logon regression test.

ironfix-engine/src/acceptor.rs
```diff
@@ -0,0 +1,504 @@
...
+        let session = Session::<Disconnected>::new(session_id.to_string()).accept();
+
+        let _ = stream.set_nodelay(true);
+        let codec = FixCodec::new()
+            .with_max_message_size(self.config.max_message_size)
+            .with_checksum_validation(self.config.validate_checksum);
+        let mut framed = Framed::new(stream, codec);
+
+        let sequences = match self.initial_sequences {
+            Some((sender, target)) if !self.config.reset_on_logon => {
```
joaquinbejar, line 327: 🔴 **Critical**: local `reset_on_logon` discards the configured continuity seed before the inbound flag is read. If the peer omits `141=Y`, this side silently starts at 1 and replies without a reset flag, reusing sender numbers and misclassifying inbound continuity. Always seed from `initial_sequences`, then reset only after a validated inbound `141=Y`; cover the local-flag true and inbound-flag absent case.

ironfix-engine/src/acceptor.rs
joaquinbejar: 🟠 **Warning**: `to_admin` mutates a decoded copy, but line 457 sends the original `reply`, so every callback edit is discarded despite the public callback contract promising modification of outgoing admin messages. Re-encode and send the mutated message, or route this through the shared callback-aware admin sender, and assert a callback-added Logon field appears on the wire.

ironfix-engine/src/acceptor.rs
```diff
@@ -0,0 +1,504 @@
...
+                {
+                    let _ = framed.send(logout).await;
+                }
+                let _ = session.reject_logon();
+                return Err(EngineError::LogonRejected {
+                    reason: reason.text,
+                });
+            }
+
+            // Honor the heartbeat interval the initiator requested (HeartBtInt,
```
joaquinbejar, line 571: 🟠 **Warning**: the handshake reaches heartbeat negotiation without validating required `EncryptMethod(98)`. A missing, malformed, or nonzero value establishes a plaintext session even though this engine only supports `98=0`. Require and strictly parse tag 98 before authentication succeeds, then reject unsupported values with typed tests.

ironfix-example/examples/fix44_server.rs
joaquinbejar, line 199: 🟠 **Warning**: `serve` has already spawned the reactor before the `Connection` is installed on line 161. A peer can pipeline `NewOrderSingle` immediately after its Logon, reach `from_app` while the slot is still empty, and have its `ExecutionReport` silently dropped. Create a response queue or readiness synchronization before the handshake and add a pipelined-order regression.

joaquinbejar (commented):
Risk is critical. The Acceptor integration and reactor extraction are substantial, but the current head permits a peer-triggered abort, duplicate live sequence streams, and silent continuity loss, with additional handshake, callback, and example races. Earlier blocking findings already recorded on the lower stack also remain unresolved. Effective verdict REQUEST_CHANGES. Submitted as COMMENT because this pull request is self-authored.

ironfix-engine/src/acceptor.rs
joaquinbejar: Fixed in 1990dcd — the Acceptor now holds a shared admission set and serve() claims an RAII AdmissionGuard once identity is validated; a second concurrent Logon for the same SessionId is refused with a Logout instead of forking into a second independent SequenceManager at seq 1. The guard rides the reactor task and frees the slot on every close path, so a genuine reconnect still works. The reconnect-continuity half is the documented no-store as-built gap — continuity is the consumer's job via with_initial_sequences.

ironfix-engine/src/acceptor.rs
joaquinbejar: Fixed in 1990dcd — the acceptor handshake now compares inbound BeginString(8) to the configured version and requires EncryptMethod(98)=0, logging out and refusing the Logon on either mismatch instead of letting a cross-version or would-be-encrypting peer reach Active. 1137/DefaultApplVerID isn't validated here (ApplVerID is negotiated via 1128 today). As you noted the Initiator's ack path shares the same omission of all three checks; extending it there is the shared follow-up.

ironfix-engine/src/acceptor.rs
joaquinbejar: Fixed in 1990dcd — the acceptor now reads MsgSeqNum(34) via wire::header_seq_num, which returns None unless 34 is in the standard header, so a message carrying 34 only in the body is rejected as missing the required header field rather than accepted on a body-only value. This extends the positional contract the identity check already used to tag 34. The body-only fix covers the acceptor handshake; the same positional gap on the shared in-session reactor path (initiator.rs uses the same get_field_as(34)) is tracked as a follow-up.

ironfix-engine/src/acceptor.rs
joaquinbejar: Fixed in 1990dcd — the acceptor now bounds HeartBtInt(108) with wire::parse_heartbeat_interval before any HeartbeatManager is built: a missing, non-numeric, or over-range value (one-day ceiling) fails the handshake with a Logout instead of silently falling back to local config or, at 108=u64::MAX, building a Duration::from_secs(u64::MAX) whose first tick overflows interval+grace and aborts the process under panic=abort. That closes the single-field remote DoS. The Initiator has the same unbounded from_secs and is the shared follow-up.

ironfix-engine/src/acceptor.rs
joaquinbejar: Fixed in 1990dcd — the reset decision is now `inbound_reset || config.reset_on_logon`, and the ack carries 141=Y whenever either drives the reset, so a locally-configured reset no longer resets the counters while acking without the flag and desyncing the peer. The MsgSeqNum=1 requirement is bound only to an inbound-declared reset now, not a purely local one, matching the with_initial_sequences doc.

ironfix-engine/src/acceptor.rs
joaquinbejar: Fixed in 1990dcd — the acceptor now sends the bytes re-encoded from the OwnedMessage the callback saw (BytesMut::from(owned.as_bytes())), not the pre-callback frame, so to_admin is authoritative over what reaches the wire. It's byte-identical today because core's OwnedMessage exposes no mutators yet; the structure is what keeps the callback in charge once it can. Making OwnedMessage mutable, and applying the same pattern on the Initiator handshake and the shared reactor's send_admin, is the separate follow-up.

ironfix-example/examples/fix44_server.rs
joaquinbejar: Fixed in 1990dcd — the example now publishes the Connection on a watch channel and from_app subscribes and awaits it, so a NewOrderSingle that races ahead of set_connection finds its handle instead of dropping the reply. It does surface a serve() API-shape point (the reply path can't be installed before inbound dispatch starts) that's worth a real engine-level fix; the watch handle is the example-level answer.

ironfix-example/examples/fix44_server.rs
joaquinbejar: Fixed in 1990dcd — from_app now builds the ExecutionReport inline (copying the values out of the borrowed message) but sends it from a spawned task instead of awaiting Connection::send on the reactor task itself, so it can't block waiting for capacity on the very command queue the reactor is the sole consumer of. It's a real shared-reactor callback-threading hazard that affects the Initiator's from_app too; the example shows the off-reactor send pattern.

ironfix-engine/src/acceptor.rs
joaquinbejar: Fixed in 1990dcd — same overflow, verified end to end: 108=u64::MAX built Duration::from_secs(u64::MAX), and should_send_test_request computing interval+1s on the first tick overflowed and aborted under panic=abort. parse_heartbeat_interval now clamps 108 to a one-day ceiling and fails the handshake on an over-range value before any manager is built. Shared with the Initiator.

ironfix-engine/src/acceptor.rs
joaquinbejar: Fixed in 1990dcd — same admission fix: the Acceptor now carries a shared admission set, so two overlapping valid Logons for one SessionId no longer each build an independent SequenceManager and both reach Active at seq 1. The second is refused with a Logout, and there's a duplicate-Logon test now. A certification desk rejecting two concurrent connections for one SessionId was the right call to encode.

ironfix-engine/src/acceptor.rs
joaquinbejar: Fixed in 1990dcd — same reset fix: the reset is `inbound_reset || config.reset_on_logon` and the ack carries 141=Y whenever either drives it, so a local reset_on_logon no longer resets to 1 without signalling the peer. The Initiator already signalled via outbound 141=Y; the acceptor does now too.

ironfix-engine/src/acceptor.rs
joaquinbejar: Fixed in 1990dcd — same to_admin fix: the reply is re-encoded from the callback's OwnedMessage before sending, so the callback edit reaches the wire. Byte-identical until core makes OwnedMessage mutable, and applying the same pattern on initiator.rs:259/449 and the reactor's send_admin is the shared follow-up.

ironfix-engine/src/acceptor.rs
joaquinbejar: Fixed in 1990dcd — the acceptor now requires EncryptMethod(98)=0 (None) on the inbound Logon and refuses a missing, non-numeric, or non-zero value with a Logout, instead of establishing a plaintext session while the peer might believe encryption was negotiated. IronFix only ever stamps 98=0 outbound, so 0 is the only method it can honestly honour. Shared with the Initiator, which is unvalidated on the same field.

ironfix-example/examples/fix44_server.rs
joaquinbejar: Fixed in 1990dcd — same race fix: the Connection is published on a watch channel and from_app awaits it, so a pipelined order that reaches from_app before set_connection still finds its handle instead of having its ExecutionReport dropped.

Issue #52, opened 2026-07-22
ironfix-engine: read the Logon-ack MsgSeqNum(34) positionally on the initiator path


joaquinbejar:
## Overview

The Initiator's Logon-**acknowledgement** MsgSeqNum read is the one remaining
place in `ironfix-engine` that reads `MsgSeqNum(34)` non-positionally, so a Logon
ack carrying a body-only `34` is accepted as the ack sequence number. Its two
siblings enforce the positional standard-header contract; the initiator ack does
not, so the same wire input is handled three different ways depending on role and
phase.

## Context

Surfaced by the protocol-reviewer of the #50 fix. After #50, positional reads of
tag 34 are enforced in two of three places:

- Acceptor handshake — `wire::header_seq_num` (`ironfix-engine/src/acceptor.rs:445`,
  tested by `test_acceptor_rejects_body_only_seq_num`).
- Shared in-session reactor — `wire::header_seq_num` (`ironfix-engine/src/reactor.rs`,
  #50).
- **Initiator Logon ack — still non-positional:** `initiator.rs:643` does
  `let ack_seq: u64 = raw.get_field_as(34)?;`, and `get_field_as` → `get_field`
  returns the *first* field with that tag regardless of position
  (`ironfix-core/src/message.rs:605,628`).

`doc/fix_operations.md:66-68` requires the ack's identity fields in the standard
header and the acceptor mirror (`:126-127`) requires the same for 34; the
initiator ack is silent by omission. Only reachable from a non-conformant
counterparty during the ack, so severity is low — but it undercuts the positional
consistency #50 establishes.

## Tasks

- [ ] Gate the initiator Logon-ack `ack_seq` through `wire::header_seq_num(&raw)`,
      returning `EngineError::Sequence(...)` on `None` (mirroring
      `acceptor.rs:445-453`)
- [ ] A Logon ack carrying `34` only after the body fails the handshake

## Acceptance criteria

- [ ] An initiator-handshake body-only-`34` test parallels
      `test_acceptor_rejects_body_only_seq_num`
- [ ] A well-formed ack with `34` in the standard header is unaffected
- [ ] `cargo test --workspace` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean

Issue #51, opened 2026-07-22
ironfix-fast: nullable i64::MAX domain excluded from the signed codec


joaquinbejar:
## Overview

`ironfix-fast`'s nullable **signed** integer codec has the same domain gap that
#48 just fixed for the unsigned side: `encode_nullable_int`/`decode_nullable_int`
bias only the non-negative half by one, so the representable domain is
`i64::MIN..=i64::MAX - 1` and `Some(i64::MAX)` is rejected as `IntegerOverflow`.
A conformant peer sending the biased `i64::MAX` value cannot be decoded.

## Context

Surfaced by the codec-expert review of the #48 fix. The unsigned form now carries
the biased value through `u128` (`ironfix-fast/src/decoder.rs` `decode_uint_u128`,
`ironfix-fast/src/encoder.rs` `encode_uint_u128`) so `Some(u64::MAX)` round-trips.
The signed form (`encode_nullable_int` / `decode_nullable_int`) still uses the
`i64` path and rejects `Some(i64::MAX)`; the encoder doc even points at "the
fallible signed encoder" as the remaining gap. Impact is low — IronFix is
self-consistent (encoder and decoder agree) and `i64::MAX` is never a real
market-data value — but full FAST conformance would decode the biased value
through `i128`, mirroring #48.

## Tasks

- [ ] Decode/encode the biased nullable signed value through `i128` so
      `Some(i64::MAX)` round-trips, keeping encoder and decoder in step
- [ ] Preserve the `MAX_INT_ENCODED_LEN` ceiling so a hostile over-long encoding
      stays a bounded `IntegerOverflow`, never a panic or unbounded read

## Acceptance criteria

- [ ] `Some(i64::MAX)` round-trips through the nullable signed codec
- [ ] A hostile over-long encoding still errors as `IntegerOverflow`
- [ ] Non-nullable `encode_int`/`decode_int` (`i64`) are unchanged
- [ ] `cargo test --workspace` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean
- [ ] Public-surface impact stated (any `pub` change is a semver event)

PR #53, opened 2026-07-22
fix: three conformance-edge fixes (Closes #48, #49, #50)

### LINKED ISSUE #48: ironfix-fast: nullable u64::MAX domain and DictionaryValue cross-variant accessors
## Overview

Two low-impact FAST conformance edges surfaced by the 2026-07-22 review round, both verified real but deliberately deferred from the review fixes because they matter only for full conformance or for a layer that does not exist yet.

### 1. Nullable unsigned domain excludes `u64::MAX`
`ironfix-fast` encodes/decodes a nullable unsigned integer with a bias of one, so the representable domain is `0..=u64::MAX - 1`; a conformant peer sending `u64::MAX` (biased to a 65-bit value) is rejected as `IntegerOverflow`. IronFix is self-consistent (the encoder also refuses it) and `u64::MAX` is never a real market-data value, so impact is low — but full conformance would decode the biased value through `u128`.

### 2. `DictionaryValue` cross-variant accessors return `None`
`DictionaryValue::Int(42).as_u64()` and `UInt(42).as_i64()` both return `None`, so a Copy or Increment operator that stored a value under one variant cannot read it back under the other. There is no operator-application layer yet, so nothing is masked today — this becomes a real interoperability trap only when the template-driven decode layer (issue #13's remaining scope) is built, and should be resolved at that typed template seam.

## Tasks

- [ ] Decode a biased nullable unsigned through `u128` so `Some(u64::MAX)` round-trips, keeping the encoder in step
- [ ] Resolve `DictionaryValue`'s cross-variant reads when the template-application layer is implemented

## Acceptance criteria

- [ ] `Some(u64::MAX)` round-trips through the nullable unsigned codec
- [ ] The cross-variant accessor behaviour is decided and tested at the template seam
- [ ] `cargo test --workspace` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean

### LINKED ISSUE #49: ironfix-core: is_user_defined(>=5000) misclassifies assigned 40000+ tags
## Overview

A pre-existing core-model overlap surfaced (and verified) during the 2026-07-22 review round: `ironfix-core`'s user-defined tag predicate overlaps the assigned high-tag range, so a standard high-numbered tag is misclassified as user-defined.

## Context

`ironfix-core`'s `FieldTag::is_user_defined` treats any tag `>= 5000` as user-defined (the FIX 5000–9999 bilateral range). But FIX also assigns standard tags in the 40000+ Extension Pack range, so a legitimately-assigned 40000+ tag is currently misclassified as user-defined. This is a `ironfix-core` field-model issue, independent of the tag=value decoder (which is a byte scanner and does not consult this predicate on the hot path).

## Tasks

- [ ] Model the user-defined ranges precisely: 5000–9999 (bilateral) is user-defined; assigned Extension-Pack tags at 40000+ are standard, not user-defined
- [ ] Decide the predicate's contract for the gaps and document it against the FIX specification

## Acceptance criteria

- [ ] `is_user_defined` returns the correct classification for a 5000-range tag, a 40000+ assigned tag, and a standard low tag
- [ ] Tests cover each range boundary
- [ ] Public-surface impact stated (any `pub` change is a semver event)

### LINKED ISSUE #50: ironfix-engine: read MsgSeqNum(34) positionally on the shared in-session reactor path
## Overview

The inbound reactor reads `MsgSeqNum(34)` from anywhere in the decoded message rather than requiring it in the standard header, so a body-placed 34 reads as present instead of a missing header field. Surfaced by a second review pass on the Acceptor (#47).

## Context

The Acceptor's own handshake was hardened to read 34 positionally (`wire::header_seq_num`, matching the standard-header check the CompID/SendingTime paths already use). But the shared in-session reactor path (`reactor.rs`, around the per-frame MsgSeqNum extraction) still uses a whole-message `get_field_as(34)`, so the same positional gap remains for in-session traffic on both the Initiator and the Acceptor, which share that reactor.

Low severity: a first-match 34 present in the header is still read correctly; the gap is that a message carrying 34 only in the body is treated as "present" rather than "missing a required header field". The binding spec (`doc/fix_operations.md`) does not assign a positional reject reason for 34, so this is inbound-acceptance hardening, not a conformance break.

## Tasks

- [ ] Read `MsgSeqNum(34)` from the standard header in the shared reactor path, reusing the `in_standard_header` positional guard
- [ ] A message with 34 only after the body is treated as a missing header field

## Acceptance criteria

- [ ] A frame carrying 34 only in the body is rejected/handled as missing on the shared in-session path, for both Initiator and Acceptor
- [ ] `cargo test --workspace` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean

joaquinbejar:
## Summary

Three small, independently-verified conformance edges from the 2026-07-22 review
round, bundled into one PR because they touch disjoint crates and share no
interface.

Closes #48
Closes #49
Closes #50

No breaking API change: one additive `pub const`, two new private helpers, and a
private-module fix. The workspace stays at unreleased `0.4.0`.

Two drive-by CI repairs ride along, both pre-existing breakage that earlier
`--admin` merges bypassed and this (first non-admin) PR surfaced:

- **`docs` job:** `cargo doc -D warnings` (stricter than the local `make doc`
  clippy gate) flagged broken intra-doc links in `ironfix-derive` and
  `ironfix-codegen` — both link into `ironfix_core`, which neither crate depends
  on directly. Demoted to plain code spans so `cargo doc --workspace` passes.
- **`smoke` job (stale asserts):** `examples.yml` grepped client logs for
  `Logon OK` / `Received:`, but #45 rewrote the example log messages to
  `logon accepted` / `market data`. The two greps now match the examples' actual
  output.
- **`smoke` job (real bug):** the Acceptor-based `fix44_server` (#47) never
  delivered its `ExecutionReport` reply — it published the `Connection` over a
  `watch` whose initial receiver is dropped, and `send` silently discards the
  value when it wins the race against `from_app`'s `subscribe` (the common case),
  so the client timed out with "no reply within 10s". `send_replace` stores the
  handle unconditionally. Verified by running the fix44 pair 4/4.
- **`build` (Docker) jobs:** the example image builds passed `cargo build
  --locked`, but `Cargo.lock` is git-ignored and excluded from the build context,
  so `--locked` had no lock file to honour — the images built only while the GHA
  cache held a compatible lock, which the `0.4.0` bumps invalidated. `--locked`
  is dropped so the images resolve dependencies fresh.

## #49 — `ironfix-core`: bound `is_user_defined` to 5000–9999

`FieldTag::is_user_defined` returned `true` for any tag `>= 5000`, so an assigned
high tag (40000+) was misclassified as user-defined. FIX defines **two disjoint
bilateral user-defined ranges**: **5000–9999**, and **20000–39999** (approved by
the GTC in December 2009 once the first filled up). Tags 10000–19999 are
internal-use and tags at or above 40000 are reserved, so neither is user-defined.

- Adds `USER_DEFINED_TAG_MAX` (9999) plus `USER_DEFINED_EXT_TAG_MIN` (20000) /
  `USER_DEFINED_EXT_TAG_MAX` (39999), and makes the predicate the union of both
  ranges.
- Boundary tests: 4999→false, 5000→true, 9999→true, 10000→false, 19999→false,
  20000→true, 39999→true, 40000→false.

## #48 — `ironfix-fast`: nullable unsigned codec now represents `u64::MAX`

**Part 1 (implemented).** The nullable unsigned codec biases by one, so
`Some(u64::MAX)` biased to `2^64`, which the `u64` encoder rejected as
`IntegerOverflow` and the `u64` decoder could not represent — the domain was
`0..=u64::MAX-1`. The biased value now travels through `u128` on both sides
(private `encode_uint_u128` / `decode_uint_u128`), so `Some(u64::MAX)`
round-trips while the plain `u64` `encode_uint`/`decode_uint` API is unchanged.
Values that fit `u64` encode to byte-identical frames as before — only the
previously-unrepresentable `Some(u64::MAX)` is new. The `MAX_INT_ENCODED_LEN`
ceiling still bounds the decode, so a hostile over-long encoding is a typed
`IntegerOverflow`, never a panic or an unbounded read.

**Part 2 (already done).** The issue's second finding — `DictionaryValue`
cross-variant accessors returning `None` — was already resolved on `main` (by
#40): `as_i64`/`as_u64` cross-read with exact-cast guards, and existing tests
pin it (`Int(42).as_u64() == Some(42)`, `UInt(42).as_i64() == Some(42)`, plus the
negative/out-of-range guards). No change was needed; the decision is test-pinned.

## #50 — `ironfix-engine`: read MsgSeqNum(34) positionally in the shared reactor

The shared in-session reactor read `MsgSeqNum(34)` with a whole-message
`get_field_str`, so a `34` placed only after the body read as present instead of
a missing header field. It now uses `wire::header_seq_num`, the same positional
guard the Acceptor handshake and the CompID/SendingTime paths use, so a body-only
`34` is treated as missing. Both `Initiator` and `Acceptor` share this reactor,
so the gap closes for in-session traffic on both. New test
`test_frame_with_body_only_msg_seq_num_is_dropped` drives a body-only-`34` frame
and asserts it is dropped without disturbing sequence state.

The spec (`doc/fix_operations.md`) assigns no positional reject reason for tag 34,
so this is inbound-acceptance hardening (silent drop, no sequence advance), not a
conformance break.

- `ironfix-core`: **adds** `USER_DEFINED_TAG_MAX`, `USER_DEFINED_EXT_TAG_MIN`,
  `USER_DEFINED_EXT_TAG_MAX` (all `pub const`, minor). No removed or renamed
  public item.
- `ironfix-fast`: behaviour change only (nullable now accepts `u64::MAX`); the two
  new helpers are private.
- `ironfix-engine`: `reactor.rs` is a private module; no public surface change.

## Tests

```
cargo test --workspace                                    # all pass
make doc                                                  # clean (clippy -W missing-docs)
```

New coverage: core tag-range boundaries; `Some(u64::MAX)` round-trip and a bounded
over-long-encoding `IntegerOverflow`; and the body-only-`34` reactor drop.
No performance claim is made — the repo has no criterion baseline.

## Review

`architect`, `codec-expert` and `protocol-reviewer` all signed off. Two review
notes were applied: the crate root now re-exports `USER_DEFINED_TAG_MAX`
symmetrically with `USER_DEFINED_TAG_MIN`, and the nullable debias uses
`checked_sub`. Two adjacent same-class gaps found in review are out of scope here
and tracked as follow-ups:

- **#51** — the nullable **signed** codec has the same `i64::MAX` domain gap #48
  fixes for unsigned.
- **#52** — the Initiator's Logon-**ack** MsgSeqNum is the one remaining
  non-positional tag-34 read (the shared reactor and the Acceptor handshake both
  enforce the standard-header contract after #50).

### DIFF
--- a/ironfix-core/src/field.rs
+++ b/ironfix-core/src/field.rs
@@ -21,18 +21,27 @@ use std::str::FromStr;
 
 /// The lowest tag number FIX reserves for user-defined fields.
 ///
-/// FIX reserves 5000-9999 for user-defined fields, so 4999 is the highest
-/// standard tag and 5000 the first user-defined one.
+/// FIX reserves 5000-9999 for bilateral user-defined fields, so 4999 is the
+/// highest standard tag below the range and 5000 the first user-defined one.
 pub const USER_DEFINED_TAG_MIN: u32 = 5000;
+
+/// The highest tag number in the FIX user-defined range.
+///
+/// The user-defined range is exactly 5000-9999. Tags at or above 10000 —
+/// including the Extension Pack tags assigned at 40000+ — are *not*
+/// user-defined; they are assigned by the specification, and the loaded
+/// dictionary (not this type) is what says what an individual high tag means.
+pub const USER_DEFINED_TAG_MAX: u32 = 9999;
 
 /// FIX field tag number.
 ///
 /// Tags are positive integers that identify fields within a FIX message.
-/// Standard tags are defined in the FIX specification (1-4999), while FIX
-/// reserves 5000-9999 for user-defined fields. Tags at or above
-/// [`USER_DEFINED_TAG_MIN`] are treated as user-defined; venues do assign
-/// beyond 9999 in practice, and a dictionary — not this type — is what says
-/// whether a given tag is known.
+/// Standard tags are defined in the FIX specification (1-4999), FIX reserves
+/// the bilateral user-defined range 5000-9999, and the specification also
+/// assigns tags above 9999 (the Extension Pack range at 40000+). Only
+/// [`USER_DEFINED_TAG_MIN`]..=[`USER_DEFINED_TAG_MAX`] is user-defined; every
+/// other value, high ones included, is standard or unassigned, and a
+/// dictionary — not this type — is what says whether a given tag is known.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 #[repr(transparent)]
 #[serde(transparent)]
@@ -77,18 +86,28 @@ impl FieldTag {
         self.0 >= 1
     }
 
-    /// Returns true if this is a standard FIX tag (1-4999).
+    /// Returns true if this is a standard FIX tag below the user-defined range
+    /// (1-4999).
+    ///
+    /// This does not cover assigned tags above 9999 (the Extension Pack range
+    /// at 40000+): those are standard too, but the dictionary — not this
+    /// predicate — classifies an individual high tag.
     #[inline]
     #[must_use]
     pub const fn is_standard(self) -> bool {
         self.0 >= 1 && self.0 < USER_DEFINED_TAG_MIN
     }
 
-    /// Returns true if this is a user-defined tag (5000+).
+    /// Returns true if this is a user-defined tag, i.e. in the bilateral range
+    /// [`USER_DEFINED_TAG_MIN`]..=[`USER_DEFINED_TAG_MAX`] (5000-9999).
+    ///
+    /// A tag at or above 10000 — including an assigned Extension Pack tag at
+    /// 40000+ — is deliberately *not* user-defined: it is assigned by the FIX
+    /// specification, not agreed bilaterally.
     #[inline]
     #[must_use]
     pub const fn is_user_defined(self) -> bool {
-        self.0 >= USER_DEFINED_TAG_MIN
+        self.0 >= USER_DEFINED_TAG_MIN && self.0 <= USER_DEFINED_TAG_MAX
     }
 }
 
@@ -423,6 +442,35 @@ mod tests {
         assert!(!tag.is_standard());
         assert!(tag.is_user_defined());
         assert_eq!(tag.value(), USER_DEFINED_TAG_MIN);
+    }
+
+    #[test]
+    fn test_field_tag_boundary_9999_is_user_defined() {
+        let tag = ok(FieldTag::try_new(9999), "9999 is a legal tag");
+        assert!(!tag.is_standard());
+        assert!(tag.is_user_defined());
+        assert_eq!(tag.value(), USER_DEFINED_TAG_MAX);
+    }
+
+    #[test]
+    fn test_field_tag_boundary_10000_is_not_user_defined() {
+        // The bilateral user-defined range ends at 9999; 10000 is outside it.
+        let tag = ok(FieldTag::try_new(10000), "10000 is a legal tag");
+        assert!(!tag.is_user_defined());
+        assert!(!tag.is_standard());
+    }
+
+    #[test]
+    fn test_field_tag_assigned_extension_pack_is_not_user_defined() {
+        // Extension Pack tags assigned at 40000+ are standard, not user-defined:
+        // the earlier `>= 5000` predicate misclassified them.
+        for tag_num in [40000, 40001] {
+            let tag = ok(FieldTag::try_new(tag_num), "assigned high tag is legal");
+            assert!(
+                !tag.is_user_defined(),
+                "assigned tag {tag_num} must not be user-defined"
+            );
+        }
     }
 
     #[test]
--- a/ironfix-core/src/field.rs
+++ b/ironfix-core/src/field.rs
@@ -19,29 +19,42 @@ use serde::{Deserialize, Serialize};
 use std::fmt;
 use std::str::FromStr;
 
-/// The lowest tag number FIX reserves for user-defined fields.
+/// The lowest tag number in the original FIX bilateral user-defined range.
 ///
 /// FIX reserves 5000-9999 for bilateral user-defined fields, so 4999 is the
 /// highest standard tag below the range and 5000 the first user-defined one.
 pub const USER_DEFINED_TAG_MIN: u32 = 5000;
 
-/// The highest tag number in the FIX user-defined range.
+/// The highest tag number in the original FIX bilateral user-defined range.
 ///
-/// The user-defined range is exactly 5000-9999. Tags at or above 10000 —
-/// including the Extension Pack tags assigned at 40000+ — are *not*
-/// user-defined; they are assigned by the specification, and the loaded
-/// dictionary (not this type) is what says what an individual high tag means.
+/// The first user-defined range is 5000-9999. Tags 10000-19999 are reserved
+/// for internal use within a single firm and are not the bilateral range.
 pub const USER_DEFINED_TAG_MAX: u32 = 9999;
+
+/// The lowest tag number in the extended FIX bilateral user-defined range.
+///
+/// The Global Technical Committee approved 20000-39999 as a second bilateral
+/// user-defined range in December 2009, once the 5000-9999 range filled up;
+/// tags here are used bilaterally and do not need to be registered.
+pub const USER_DEFINED_EXT_TAG_MIN: u32 = 20000;
+
+/// The highest tag number in the extended FIX bilateral user-defined range.
+///
+/// The extended user-defined range is 20000-39999. Tags at or above 40000 are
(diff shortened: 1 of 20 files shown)


ironfix-core/src/field.rs
```diff
@@ -77,18 +86,28 @@ impl FieldTag {
...
+    /// [`USER_DEFINED_TAG_MIN`]..=[`USER_DEFINED_TAG_MAX`] (5000-9999).
+    ///
+    /// A tag at or above 10000 — including an assigned Extension Pack tag at
+    /// 40000+ — is deliberately *not* user-defined: it is assigned by the FIX
+    /// specification, not agreed bilaterally.
     #[inline]
     #[must_use]
     pub const fn is_user_defined(self) -> bool {
-        self.0 >= USER_DEFINED_TAG_MIN
+        self.0 >= USER_DEFINED_TAG_MIN && self.0 <= USER_DEFINED_TAG_MAX
```
joaquinbejar: 🔴 **Critical**: FIX Latest defines two user-defined ranges, 5000–9999 and 20000–39999. This bound regresses every approved bilateral tag in the second range from true to false. Model both disjoint ranges and add boundary coverage at 19999, 20000, 39999, and 40000; 10000–19999 are internal-use and 40000–49999 are GTC-reserved.

ironfix-fast/src/decoder.rs
```diff
@@ -225,18 +267,26 @@ impl FastDecoder {
...
+        let biased = Self::decode_uint_u128(data, offset)?;
+        match biased {
             0 => Ok(None),
+            // `checked_sub` keeps the decode path free of bare arithmetic even
+            // though the `0` arm above already guarantees `biased >= 1`; the
+            // debiased value may still exceed `u64::MAX` for an over-long
+            // encoding, which the narrowing rejects as `IntegerOverflow`.
             biased => biased
                 .checked_sub(1)
+                .and_then(|value| u64::try_from(value).ok())
```
joaquinbejar, line 289: 🟠 **Warning**: the new narrowing failure is not exercised independently. The hostile vector below has no stop bit and exits at the ten-byte ceiling, while the maximum case is generated by the matching encoder. Add exact external fixtures for biased `2^64` and stopped `2^64 + 1` so this branch is proven to accept the boundary and reject the first out-of-range value.

ironfix-fast/src/encoder.rs
```diff
@@ -176,24 +176,49 @@ impl FastEncoder {
...
+    }
+
     /// Encodes a nullable unsigned integer.
     ///
     /// A nullable unsigned integer is encoded with a bias of one so that the
-    /// single stop byte `0x80` is reserved for `None`. The representable
-    /// domain is therefore `0..=u64::MAX - 1`: `Some(u64::MAX)` has no
-    /// encoding and is rejected rather than silently biased into the `None`
-    /// representation.
+    /// single stop byte `0x80` is reserved for `None`. The full `0..=u64::MAX`
```
joaquinbejar, line 203: 🟠 **Warning**: this behavior change leaves `ironfix-fast/src/lib.rs` saying that `Some(u64::MAX)` is refused with a typed error. Update the crate-level docs in this PR; the public documentation currently contradicts the implemented domain.

joaquinbejar (commented):
The FAST codec boundary and positional sequence guard are otherwise sound, but the UDF predicate drops an official bilateral range, the new narrowing branch lacks a direct malformed fixture, and the crate docs now contradict the encoder. REQUEST_CHANGES.

ironfix-core/src/field.rs
joaquinbejar: You're right, good catch. I confirmed the GTC approved 20000 to 39999 as a second bilateral user defined range in December 2009, so is_user_defined now covers both that and 5000 to 9999 through new USER_DEFINED_EXT_TAG consts, with boundary tests at 19999, 20000, 39999 and 40000.

ironfix-fast/src/decoder.rs
joaquinbejar: Agreed, the round trip and the ten byte ceiling case both sidestep the try_from branch. I added hand built stop bit fixtures, biased 2^64 decodes to Some(u64::MAX) and biased 2^64 plus 1 is IntegerOverflow, so the accepted boundary and the first out of range value are both pinned directly.

ironfix-fast/src/encoder.rs
joaquinbejar: Fixed, the ironfix-fast crate docs no longer say Some(u64::MAX) is refused and now state the nullable unsigned codec covers the full 0 to u64::MAX domain.

joaquinbejar:
Thanks for the thorough review, all three are fixed and pushed in ce09c83. is_user_defined now models both bilateral ranges 5000 to 9999 and 20000 to 39999, the u128 narrowing branch has direct external fixtures for biased 2^64 and 2^64 plus 1, and the ironfix-fast crate docs match the widened domain.

PR #54, opened 2026-07-22
fix: nullable signed i64::MAX and positional initiator ack MsgSeqNum (Closes #51, #52)

### LINKED ISSUE #51: ironfix-fast: nullable i64::MAX domain excluded from the signed codec

### LINKED ISSUE #52: ironfix-engine: read the Logon-ack MsgSeqNum(34) positionally on the initiator path

joaquinbejar:
## Summary

Two small follow-ups from the #48/#49/#50 review round, each the same class as a
fix already shipped. They touch disjoint crates (`ironfix-fast`, `ironfix-engine`)
and share no interface, but land together in one PR as requested.

Closes #51
Closes #52

No breaking API change: no `pub` item added, renamed, or reshaped. The workspace
stays at unreleased `0.4.0`.

## #51 — `ironfix-fast`: nullable **signed** codec now represents `i64::MAX`

The exact signed counterpart of the unsigned fix from #48. The nullable signed
codec biases the non-negative half by one, so `Some(i64::MAX)` biased to `2^63`,
which the `i64` encoder rejected as `IntegerOverflow` and the `i64` decoder could
not represent — the domain was `i64::MIN..=i64::MAX-1`. The biased value now
travels through `i128` on both sides (private `encode_int_i128` /
`decode_int_i128`), so `Some(i64::MAX)` round-trips while the plain `i64`
`encode_int`/`decode_int` API and the negative (unbiased) path are unchanged.
The `MAX_INT_ENCODED_LEN` ceiling still bounds the decode, so a hostile over-long
encoding is a typed `IntegerOverflow`, never a panic or an unbounded read.

New coverage: `Some(i64::MAX)` round-trip; a bounded over-long-encoding
`IntegerOverflow`; and hand-built external stop-bit fixtures pinning the `i128`
narrowing branch directly (biased `2^63` → `Some(i64::MAX)`, biased `2^63+1` →
`IntegerOverflow`).

## #52 — `ironfix-engine`: read the Logon-ack MsgSeqNum(34) positionally

The initiator read the Logon-ack sequence number with a whole-message
`get_field_as(34)`, so an ack carrying `34` only after the body was accepted —
the last non-positional tag-34 read in the engine (the Acceptor handshake and the
shared reactor already read it positionally). It now uses `wire::header_seq_num`,
and fails the handshake with `EngineError::Sequence` when `34` is absent from the
standard header, calling `on_logon_reject` first — mirroring the Acceptor's
`test_acceptor_rejects_body_only_seq_num`.

New coverage: `test_logon_ack_body_only_msg_seq_num_fails_handshake` drives a
body-only-`34` ack and asserts the handshake fails; the well-formed header-`34`
path is unchanged (the existing 80+ initiator tests cover it).

None. The two `ironfix-fast` helpers are private; `initiator.rs` is
engine-internal and `EngineError::Sequence` already exists. Behaviour changes
only (nullable signed accepts `i64::MAX`; the initiator ack fails on a body-only
`34`), which are not semver-breaking API changes.

## Verification

```
cargo test --workspace                                    # all pass
cargo doc --no-deps --workspace --all-features            # clean (-D warnings)
```

No performance claim is made — the repo has no criterion baseline.

### DIFF
--- a/doc/fix_operations.md
+++ b/doc/fix_operations.md
@@ -106,11 +106,11 @@ and stops at the first failure:
    — its initiator would set 1, its acceptor 2 — and IronFix matches the
    acceptor. The mismatch self-heals: the peer sees a gap at 2, requests a
    resend of 1, and receives a GapFill that resynchronises it.
-7. **`MsgSeqNum` (34).** Read positionally: 34 must appear in the standard
-   header, the same contract steps 2 and 5 enforce, so a 34 that occurs only
-   after the body is treated as missing and fails the handshake. When present in
-   the header, too low fails the handshake and a gap completes the handshake and
-   immediately issues a ResendRequest.
+7. **`MsgSeqNum` (34).** The header 34 is read positionally, *before* identity —
+   the same standard-header contract step 2 applies to the CompIDs — so a 34 that
+   occurs only after the body is treated as missing and fails the handshake. The
+   value read there is then validated at this step: too low fails the handshake,
+   and a gap completes the handshake and immediately issues a ResendRequest.
 
 The same identity and `SendingTime` checks (steps 2 and 3) run on every inbound
 frame once the session is established; there a failure produces the Reject and
--- a/doc/fix_operations.md
+++ b/doc/fix_operations.md
@@ -106,8 +106,11 @@ and stops at the first failure:
    — its initiator would set 1, its acceptor 2 — and IronFix matches the
    acceptor. The mismatch self-heals: the peer sees a gap at 2, requests a
    resend of 1, and receives a GapFill that resynchronises it.
-7. **`MsgSeqNum` (34).** Too low fails the handshake; a gap completes the
-   handshake and immediately issues a ResendRequest.
+7. **`MsgSeqNum` (34).** Read positionally: 34 must appear in the standard
+   header, the same contract steps 2 and 5 enforce, so a 34 that occurs only
+   after the body is treated as missing and fails the handshake. When present in
+   the header, too low fails the handshake and a gap completes the handshake and
+   immediately issues a ResendRequest.
 
 The same identity and `SendingTime` checks (steps 2 and 3) run on every inbound
 frame once the session is established; there a failure produces the Reject and
--- a/ironfix-fast/src/decoder.rs
+++ b/ironfix-fast/src/decoder.rs
@@ -379,10 +379,13 @@ impl FastDecoder {
         }
 
         // Positive values are biased by one; `2^63` debiases to i64::MAX, and
-        // anything larger overflows i64.
-        i64::try_from(raw - 1)
+        // anything larger overflows i64. `checked_sub` keeps the decode path
+        // free of bare arithmetic, matching the unsigned sibling, though the
+        // `raw > 0` branch already guarantees no underflow.
+        raw.checked_sub(1)
+            .and_then(|value| i64::try_from(value).ok())
             .map(Some)
-            .map_err(|_| FastError::IntegerOverflow)
+            .ok_or(FastError::IntegerOverflow)
     }
 
     /// Decodes a nullable ASCII string.
--- a/ironfix-fast/src/decoder.rs
+++ b/ironfix-fast/src/decoder.rs
@@ -292,6 +292,53 @@ impl FastDecoder {
         }
     }
 
+    /// Decodes a stop-bit signed integer into an `i128`.
+    ///
+    /// Identical to [`FastDecoder::decode_int`] but accumulates in `i128`, so it
+    /// can hold the biased `2^63` a nullable signed uses to denote `i64::MAX`.
+    /// The same [`MAX_INT_ENCODED_LEN`] ceiling bounds the read, so a hostile
+    /// over-long encoding is [`FastError::IntegerOverflow`] rather than an
+    /// unbounded loop.
+    ///
+    /// # Errors
+    /// [`FastError::UnexpectedEof`] if the input ends mid-value, or
+    /// [`FastError::IntegerOverflow`] if the encoding exceeds
+    /// [`MAX_INT_ENCODED_LEN`] bytes.
+    fn decode_int_i128(data: &[u8], offset: &mut usize) -> Result<i128, FastError> {
+        /// Value of one payload byte position.
+        const RADIX: i128 = 1 << PAYLOAD_BITS;
+
+        let first_byte = *data.get(*offset).ok_or(FastError::UnexpectedEof)?;
+        let negative = (first_byte & SIGN_BIT) != 0;
+
+        // Sign extension: a negative value starts from all-ones so the
+        // accumulated two's complement stays negative.
+        let mut result: i128 = if negative { -1 } else { 0 };
+        let mut consumed: usize = 0;
+
+        loop {
+            if consumed == MAX_INT_ENCODED_LEN {
+                return Err(FastError::IntegerOverflow);
+            }
+            let byte = read_byte(data, offset)?;
+            consumed += 1;
+
+            // `checked_mul` rejects any value that would lose significant bits.
+            // The product is a multiple of `RADIX`, so its low seven bits are
+            // zero and the `|` below is exact.
+            result = result
+                .checked_mul(RADIX)
+                .ok_or(FastError::IntegerOverflow)?
+                | i128::from(byte & PAYLOAD_MASK);
+
+            if byte & STOP_BIT != 0 {
+                break;
+            }
+        }
+
+        Ok(result)
+    }
+
     /// Decodes a nullable signed integer.
     ///
     /// FAST biases only the non-negative half of the range: zero is NULL, a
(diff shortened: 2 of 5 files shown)


ironfix-fast/src/decoder.rs
```diff
@@ -292,6 +292,53 @@ impl FastDecoder {
...
+        // accumulated two's complement stays negative.
+        let mut result: i128 = if negative { -1 } else { 0 };
+        let mut consumed: usize = 0;
+
+        loop {
+            if consumed == MAX_INT_ENCODED_LEN {
+                return Err(FastError::IntegerOverflow);
+            }
+            let byte = read_byte(data, offset)?;
+            consumed += 1;
```
joaquinbejar: 🟠 **Warning**: this new untrusted-input length counter uses bare addition. Even though the ten-byte guard bounds it, the repository definition of done requires checked arithmetic on decoder lengths and offsets. Use checked_add(1).ok_or(FastError::IntegerOverflow)? here.

ironfix-fast/src/encoder.rs
```diff
@@ -226,26 +226,63 @@ impl FastEncoder {
...
+    /// Never returns an error; the `Result` is kept for API stability and to
+    /// match the fallible ASCII and byte-vector encoders.
     pub fn encode_nullable_int(&mut self, value: Option<i64>) -> Result<(), FastError> {
         match value {
             Some(v) if v < 0 => self.encode_int(v),
             Some(v) => {
-                let biased = v.checked_add(1).ok_or(FastError::IntegerOverflow)?;
-                self.encode_int(biased);
+                // The biased value is at most `2^63`, which fits `i128`, not `i64`.
+                let biased = i128::from(v) + 1;
```
joaquinbejar: 🟠 **Warning**: the nullable bias is another arithmetic edge introduced by this change, but it uses bare addition. Convert it to checked_add(1).ok_or(FastError::IntegerOverflow)? so the implementation matches the binding checked-arithmetic rule and the PR claim that intermediate arithmetic is checked.

ironfix-fast/src/decoder.rs
```diff
@@ -313,20 +362,28 @@ impl FastDecoder {
...
         if raw == 0 {
             return Ok(None);
         }
 
         if raw < 0 {
-            // Negative values carry no bias.
-            return Ok(Some(raw));
+            // Negative values carry no bias; an over-long encoding could still
+            // land below i64::MIN, which the narrowing rejects.
+            return i64::try_from(raw)
```
joaquinbejar, line 378: 🟠 **Warning**: this new negative narrowing branch is a distinct malformed-input exit, but the independent fixtures cover only positive overflow. Add a fully stopped ten-byte value below i64::MIN (for example [0x7e, 0x7f, 0x7f, 0x7f, 0x7f, 0x7f, 0x7f, 0x7f, 0x7f, 0xff]) and assert FastError::IntegerOverflow. Also pin truncated input through decode_nullable_int, since the existing EOF tests exercise the old decode_int helper rather than this new i128 path.

doc/fix_operations.md
```diff
@@ -106,8 +106,11 @@ and stops at the first failure:
    — its initiator would set 1, its acceptor 2 — and IronFix matches the
    acceptor. The mismatch self-heals: the peer sees a gap at 2, requests a
    resend of 1, and receives a GapFill that resynchronises it.
-7. **`MsgSeqNum` (34).** Too low fails the handshake; a gap completes the
-   handshake and immediately issues a ResendRequest.
+7. **`MsgSeqNum` (34).** Read positionally: 34 must appear in the standard
+   header, the same contract steps 2 and 5 enforce, so a 34 that occurs only
```
joaquinbejar: 🟠 **Warning**: this section says validation runs in numbered order, but header_seq_num is evaluated before identity, not at step 7. The cross-reference is also false: step 5 is HeartBtInt (108), a Logon body field that does not enforce standard-header placement. Move or describe this positional check at its real execution point and refer only to step 2 or to the acceptor flow.

joaquinbejar (commented):
Decision: REQUEST_CHANGES. The FAST widening and positional Logon-ack lookup are otherwise conformant, and all affected tests and doctests pass locally, but two new arithmetic sites violate the repository checked-arithmetic rule, the negative narrowing error path lacks its required adversarial fixture, and the binding operations document is inaccurate about both validation order and HeartBtInt placement. No DAG, public-API, semver, allocation, or unbounded-scan regression was found.

ironfix-fast/src/decoder.rs
joaquinbejar: Done, decode_int_i128's counter now uses checked_add even though the ten byte ceiling already bounds it, so it matches the rule.

ironfix-fast/src/encoder.rs
joaquinbejar: Fixed, the i128 bias uses checked_add now. It cannot actually overflow for an i64 input, but the arithmetic is explicit rather than assumed.

ironfix-fast/src/decoder.rs
joaquinbejar: Good call, I added both. A fully stopped ten byte value below i64::MIN exercises the negative narrowing branch as IntegerOverflow, and a truncated single byte pins UnexpectedEof through the new i128 path.

doc/fix_operations.md
joaquinbejar: You're right, step 5 is HeartBtInt and does not enforce header placement. Step 7 now references only step 2 and says the positional 34 read runs before identity, with the too low and gap validation happening at this step.

joaquinbejar:
Thanks, all four addressed and pushed in e6e3914. The two new arithmetic sites use checked_add now, the negative narrowing and truncated i128 paths have their own adversarial fixtures, and doc step 7 no longer cites step 5 and notes the positional 34 read runs before identity.

